Issue #25, opened 2021-10-28
Add book-title reconciliation mechanism


rgaudin:
Required for #24, a mechanism to find the appropriate Title for a book's metadata, if there is such a title in the database.

rgaudin:
First of all, book-title reconciliation should happen during book creation. If it fails (not likely), then we'll end up with dangling books that will need to be assigned Titles to manually.

Titles are identified by an `ident` property which we want to mimic [Zimfarm's recipes](https://farm.openzim.org/recipes).

Should we receive a *valid* (see format bellow) `recipe` property in the endpoint, we should use it. I will check with @kelson42 whether we want Zimfarm to submit its recipe name to book-add endpoint.

The title/Zimfarm-recipe format looks like: `{project}_{lang}[_{classifier}]`

- `project` is either an identified project name like `wikipedia`, `wikibooks`, or `gutenberg` or a domain-name (`stackoverflow.com`), depending on the source.
- `lang` is the ISO 639-1 (2-letters long) language code of the content. ⚠️ this is different to the 3-letters ISO 639-3 used in the `Language` metadata. zimscraperlib can convert it though. Multi-language ZIMs may either use the most important language or `mul` (which is not ISO 639-1 valid).
- `classifier` is optional and can represent different things:
    - a `selection`. For instance, on Wikipedia ZIMs, we have the `all` selection meaning all articles, but also `top` or `football` or `movies`, etc.
    - a `flavour`. This is a tailored version of a ZIM for a specific use or audience. We often use `maxi` or `nopic` or `mini` but this is flexible. Flavour *may be found* in `Flavour` metadata.
    - `selection` and `flavour` can be combined.

Sample existing, valid recipe names:

- `wiktionary_af`
- `stackoverflow.com_bard`
- `wikihow_ar_maxi`
- `les-fondamentaux_fr_all`
- `wikipedia_en_all_maxi`
- `french.stackexchange.com_mul`

---

This format is close to the `Name` metadata (which is mandatory) found in Books. Reconciliation will thus be from `Name` metadata to `Tilte.ident`.

I don't think we want to fallback to using filenames but will check. Those usually look like `{Name}_{period-year}-{period-month}.zim`.

When creating Titles (when we're confident there is none for this book), we'll probably reuse that name but we need to ensure the format is valid with what's above.

rgaudin:
[...]

PR #36, opened 2021-11-16
Add endpoint books-add

### LINKED ISSUE #24: Add add-book endpoint
`/books/add` API endpoint to receive Book addition requests.

This will be used by the Zimfarm initially.

Endpoint will receive all *metadata* of a Book (a ZIM file) and add it to the Database. Should there be no associated Title, a Title should be created using the book metadata as well.

anshulxyz:
Closes #24

- Add a test for the endpoint
- Replace 'uuid' with 'id' in test code
- Save Metadata through `BookMetadata`
- Save Tags through `BookTag`
- Save Language

### DIFF
--- a/backend/src/backend/main.py
+++ b/backend/src/backend/main.py
@@ -2,14 +2,26 @@
 # -*- coding: utf-8 -*-
 # vim: ai ts=4 sts=4 et sw=4 nu
 
+import base64
 import datetime
 
 from fastapi import FastAPI, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import RedirectResponse
+from zimscraperlib.i18n import find_language_names
 
 from backend import __description__, __title__, __version__
 from backend.constants import BackendConf
+from backend.models import (
+    KIND_ILLUSTRATION,
+    KIND_TEXT,
+    Book,
+    BookMetadata,
+    BookTag,
+    Language,
+    database,
+)
+from backend.schemas import BookAddSchema
 
 PREFIX = "/v1"
 
@@ -49,6 +61,54 @@ async def test(timestamp: int):
         raise HTTPException(status_code=400, detail=f"Invalid timestamp: {exc}")
 
     return {"requested_on": requested_on, "received_on": received_on}
+
+
+@database.transaction()
+@api.post("/books/add/", status_code=201)
+async def add_book(book_payload: BookAddSchema):
+    """API endpoint to receive Book addition requests and add to database"""
+
+    book = await Book.objects.create(
+        id=book_payload.id,
+        counter=book_payload.counter,
+        period=book_payload.period,
+        article_count=book_payload.article_count,
+        media_count=book_payload.media_count,
+        size=book_payload.size,
+        url=book_payload.url,
+        zimcheck=book_payload.zimcheck,
+    )
+
+    for metadata_name, value in book_payload.metadata.items():
+        if metadata_name.startswith("Illustration_"):
+            await BookMetadata.objects.create(
+                book=book.id,
+                name=metadata_name,
+                bin_value=base64.standard_b64decode(value),
+                kind=KIND_ILLUSTRATION,
+            )
+        else:
+            await BookMetadata.objects.create(
+                book=book.id,
+                name=metadata_name,
+                value=value,
+                kind=KIND_TEXT,
+            )
+
+    for tag_name in book_payload.metadata["Tags"].split(";"):
+        book_tag = await BookTag.objects.get_or_create(name=tag_name)
+        await book.tags.add(book_tag)
+
+    for lang_code in book_payload.metadata["Language"].split(","):
+        # tuple format (native, english)
+        lang_tuple = find_language_names(lang_code)
+
+        language = await Language.objects.get_or_create(
+            code=lang_code, name=lang_tuple[1], native=lang_tuple[0]
+        )
+        await book.languages.add(language)
+
+    return {"msg": "ok"}
 
 
 app.mount(PREFIX, api)
--- a/backend/src/backend/schemas.py
+++ b/backend/src/backend/schemas.py
@@ -1,0 +1,18 @@
+import datetime
+import uuid
+from typing import Any, Dict
+
+import pydantic
+from pydantic import BaseModel
+
+
+class BookAddSchema(BaseModel):
+    id: uuid.UUID
+    period: datetime.date
+    counter: Dict[str, Any]
+    article_count: int
+    media_count: int
+    size: int
+    metadata: Dict[str, Any]
+    url: pydantic.HttpUrl
+    zimcheck: Dict[str, Any]
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -30,13 +30,31 @@ async def language_fra(book_dict):
 @pytest.fixture(scope="function")
 def book_dict():
     return {
-        "uuid": uuid.uuid4().hex,
+        "id": str(uuid.uuid4()),
         "period": "2021-02-08",
         "counter": {"text/html": 1, "image/png": 1},
         "article_count": 2,
         "media_count": 1,
         "size": 2048,
         "url": "https://mirror.download.kiwix.org/zim/other/test.zim",
+        "metadata": {
+            "Counter": "text/plain=10;text/css=22;application/javascript=23;"
+            "image/png=1;text/html=2567;image/webp=719;image/svg+xml;"
+            " charset=utf-8; "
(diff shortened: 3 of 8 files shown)


backend/tests/conftest.py
```diff
@@ -30,7 +30,7 @@ async def language_fra(book_dict):
 @pytest.fixture(scope="function")
 def book_dict():
     return {
-        "uuid": uuid.uuid4().hex,
+        "id": str(uuid.uuid4()),
```
anshulxyz, line 33: The model says that we're storing the UUID format is string.

anshulxyz:
hi @rgaudin I have fixed the following (I believe), thoughts?:
> we want to create `BookMetadata` for each entry in the input `metadata` and link them to the created Book.

Working on the `Language` and `Tags` now.

backend/tests/conftest.py
rgaudin: Yes, actually we use `id` terminology in library.xml so it makes more sense like this. 👍

backend/src/backend/schemas.py
```diff
@@ -0,0 +1,22 @@
...
+    period: datetime.date
+    counter: Dict[str, Any]
+    article_count: int
+    media_count: int
+    size: int
+    metadata: Dict[str, Any]
+    url: pydantic.HttpUrl
+    zimcheck: Dict[str, Any]
+    obsolete: bool
+    languages: Optional[str]
```
rgaudin, line 20: we won't be sending this over the API. This will be retrieved from metdata.

backend/src/backend/schemas.py
```diff
@@ -0,0 +1,22 @@
...
+    tags: Optional[str]
```
rgaudin, line 21: Idem

backend/src/backend/schemas.py
```diff
@@ -0,0 +1,22 @@
...
+    title: Optional[str]
```
rgaudin, line 22: Idem

backend/src/backend/main.py
```diff
@@ -51,4 +53,28 @@ async def test(timestamp: int):
...
+        url=book.url,
+        zimcheck=book.zimcheck,
+    )
+
+    for key, value in book.metadata.items():
+        await BookMetadata.objects.create(
+            book=book.id,
+            name=key,
+            value=value,
+            kind=KIND_TEXT,
```
rgaudin, line 74: Illustrations must be filtered-out and handled differently

backend/src/backend/main.py
```diff
@@ -51,4 +53,28 @@ async def test(timestamp: int):
     return {"requested_on": requested_on, "received_on": received_on}
 
 
+@api.post("/books/add/", status_code=201)
+async def books_add(book: SchemaBook):
```
rgaudin, line 57: I'd prefer the function's name to be action-like. `add_book()` for instance. We'll probably group the *routes* into dedicated modules sometime.

I don't think using `book` as variable name is appropriate as we should reserve that for our model instances. We can use something like `payload`, `book_payload`, or `book_data`. Not sure what's best yet so I'll let you choose one that distinctive from our DB Model.

backend/src/backend/main.py
```diff
@@ -51,4 +53,28 @@ async def test(timestamp: int):
...
+    await Book.objects.create(
```
rgaudin, line 58: You may want to wrap this all into a transaction so that we don't end-up with partial data in case something goes wrong later-on as this is a multi-stage, multi-table insertion

backend/src/backend/main.py
```diff
@@ -51,4 +53,28 @@ async def test(timestamp: int):
...
+        article_count=book.article_count,
+        media_count=book.media_count,
+        size=book.size,
+        url=book.url,
+        zimcheck=book.zimcheck,
+    )
+
+    for key, value in book.metadata.items():
+        await BookMetadata.objects.create(
+            book=book.id,
```
rgaudin, line 71: I find reusing the above-created `Book` instance as param for the foreign-key to be a good practice. Ensures previous steps is all-good (easy to miss an `await` and end-up with unexpected state)

backend/src/backend/main.py
```diff
@@ -51,4 +53,28 @@ async def test(timestamp: int):
...
+            name=key,
+            value=value,
+            kind=KIND_TEXT,
+        )
+
+    return {"msg": "ok"}
```
rgaudin, line 77: We'll need to define what we want to return beside the HTTP response code. Good for now

backend/src/backend/main.py
```diff
@@ -53,27 +60,39 @@ async def test(timestamp: int):
...
+        kind = (
+            KIND_TEXT
+            if not metadata_name.startswith("Illustration_")
+            else KIND_ILLUSTRATION
+        )
         await BookMetadata.objects.create(
             book=book.id,
-            name=key,
+            name=metadata_name,
             value=value,
```
rgaudin, line 88: If you look at the model, you'll see that we store the content into different fields for binary/illustration. You want to save the decoded base64 into `bin_value`. This allows us to run text-only queries on text metadata without hassle.

backend/tests/test_stub.py
```diff
@@ -54,21 +54,38 @@ async def test_books_add_endpoint(book_dict):
...
     assert response.headers.get("Content-Type") == "application/json"
 
-    book = await Book.objects.get(id=book_dict["id"])
-    assert str(book.id) == book_dict["id"]
-    assert book.counter == book_dict["counter"]
-    assert book.period == datetime.date.fromisoformat(book_dict["period"])
-    assert book.article_count == book_dict["article_count"]
-    assert book.size == book_dict["size"]
-    assert book.zimcheck == book_dict["zimcheck"]
+    _book = await Book.objects.get(id=book_dict["id"])
```
rgaudin, line 57: python convention is to use `_` as a suffix when you don't want it to collide with another variable. so `book_` instead.

backend/tests/test_stub.py
```diff
@@ -69,13 +69,19 @@ async def test_books_add_endpoint_save_book_metadata(book_dict):
...
 
-    _metadata = await BookMetadata.objects.all(book=book_dict["id"])
-    for _ in _metadata:
-        assert _.value == book_dict["metadata"][_.name]
+    metadata_ = await BookMetadata.objects.fields("id", "metadata").all(
+        book=book_dict["id"]
+    )
+
+    assert len(metadata_) == len(book_dict["metadata"])
+    for _ in metadata_:
```
rgaudin, line 77: never use `_` if you intend to use it. `_` is a convention for stuff that is not used.

backend/src/backend/main.py
```diff
@@ -77,17 +77,20 @@ async def add_book(book_payload: BookAddSchema):
...
-            book=book.id,
-            name=metadata_name,
-            value=value,
-            kind=kind,
-        )
+        if metadata_name.startswith("Illustration_"):
+            await BookMetadata.objects.create(
+                book=book.id,
+                name=metadata_name,
+                bin_value=value,
```
rgaudin, line 84: Shouldn't that be base64 decoded ?

backend/tests/test_stub.py
anshulxyz: @rgaudin does this also apply to list/set comprehension?

backend/src/backend/main.py
rgaudin: The string is base64 encoded ; look at the tests for an example of what we want to do

backend/tests/test_stub.py
rgaudin: absolutely. You use it bellow (`_.name`). If you use it, use a properly named variable like `item` here.

`_` is used when a return values is not to be used.

```py
for _ in range(10):
    print(".")  # don't need the number returned by range here
```

anshulxyz:
> I suppose we can merge this without and you'll add that part to this endpoint when doing #25. Is that OK ?

Yes, it's okay with me. I'll finish the #25 now ASAP

anshulxyz:
I'd also like the commits to be squashed together into one for the merge, if the policy allows and you see it fit.

rgaudin (approved):
👍

rgaudin:
> I'd also like the commits to be squashed together into one for the merge, if the policy allows and you see it fit.

Yes, it's a good fit for this particular case as the changes surface is small and makes sense as a whole.

Now that it's accepted, the process would be for you to rebase and squash into the appropriate number of commits (1 here).

You can do that and force push or I'll squash tomorrow morning.

anshulxyz:
> ... squash into the appropriate number of commits ...
>You can do that and force push

Done

PR #37, opened 2021-11-29
Add title reconciliation mechanism

### LINKED ISSUE #25: Add book-title reconciliation mechanism

anshulxyz:
- Add `utils.py` and `test_utils.py`
- Add `get_ident_from_name()`

Closes #25

### DIFF
--- a/backend/src/backend/main.py
+++ b/backend/src/backend/main.py
@@ -74,8 +74,6 @@ async def add_book(book_payload: BookAddSchema):
     if ident:
         # if valid ident
         title = await Title.objects.get_or_create(ident=ident)
-    else:
-        title = None
 
     book = await Book.objects.create(
         id=book_payload.id,
@@ -86,7 +84,7 @@ async def add_book(book_payload: BookAddSchema):
         size=book_payload.size,
         url=book_payload.url,
         zimcheck=book_payload.zimcheck,
-        title=title,
+        title=title if ident else None,
     )
 
     for metadata_name, value in book_payload.metadata.items():
--- a/backend/src/backend/main.py
+++ b/backend/src/backend/main.py
@@ -4,6 +4,7 @@
 
 import base64
 import datetime
+import re
 
 from fastapi import FastAPI, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
@@ -71,6 +72,16 @@ async def test(timestamp: int):
 async def add_book(book_payload: BookAddSchema):
     """API endpoint to receive Book addition requests and add to database"""
 
+    try:
+        title = await Title.objects.get_or_create(
+            ident=get_ident_from_name(book_payload.metadata["Name"])
+        )
+    except ValueError as exc:
+        raise HTTPException(
+            status_code=400,
+            detail=f"Invalid Name metadata. Unable to build Title identifier: {exc}",
+        ) from exc
+
     book = await Book.objects.create(
         id=book_payload.id,
         counter=book_payload.counter,
@@ -81,6 +92,8 @@ async def add_book(book_payload: BookAddSchema):
         url=book_payload.url,
         zimcheck=book_payload.zimcheck,
     )
+
+    await book.update(title=title)
 
     for metadata_name, value in book_payload.metadata.items():
         if metadata_name.startswith("Illustration_"):
@@ -98,16 +111,7 @@ async def add_book(book_payload: BookAddSchema):
                 kind=KIND_TEXT,
             )
 
-    try:
-        title = await Title.objects.get_or_create(
-            ident=get_ident_from_name(book_payload.metadata["Name"])
-        )
-    except ValueError as exc:
-        raise HTTPException(status_code=400, detail=f"Exception: {exc}") from exc
-
-    await book.update(title=title)
-
-    for metadata in await BookMetadata.objects.all(book=book.id):
+    for metadata in await book.metadata.all():
         await TitleMetadata.objects.create(
             title=title.ident,
             name=metadata.name,
@@ -119,9 +123,7 @@ async def add_book(book_payload: BookAddSchema):
     for tag_name in book_payload.metadata["Tags"].split(";"):
         book_tag = await BookTag.objects.get_or_create(name=tag_name)
         await book.tags.add(book_tag)
-        if not tag_name.startswith(
-            ("_sw", "_ftindex", "_pictures", "_videos", "_details")
-        ):
+        if not re.match(r"_(sw|ftindex|pictures|videos|details):(yes|no)", tag_name):
             await title.tags.add(book_tag)
 
     for lang_code in book_payload.metadata["Language"].split(","):
--- a/backend/src/backend/main.py
+++ b/backend/src/backend/main.py
@@ -19,9 +19,11 @@ from backend.models import (
     BookMetadata,
     BookTag,
     Language,
+    Title,
     database,
 )
 from backend.schemas import BookAddSchema
+from backend.utils import get_ident_from_name
 
 PREFIX = "/v1"
 
@@ -68,6 +70,13 @@ async def test(timestamp: int):
 async def add_book(book_payload: BookAddSchema):
     """API endpoint to receive Book addition requests and add to database"""
 
+    ident = get_ident_from_name(book_payload.metadata["Name"])
+    if ident:
+        # if valid ident
+        title = await Title.objects.get_or_create(ident=ident)
+    else:
+        title = None
+
     book = await Book.objects.create(
         id=book_payload.id,
         counter=book_payload.counter,
@@ -77,6 +86,7 @@ async def add_book(book_payload: BookAddSchema):
         size=book_payload.size,
         url=book_payload.url,
         zimcheck=book_payload.zimcheck,
+        title=title,
     )
 
     for metadata_name, value in book_payload.metadata.items():
--- a/backend/src/backend/main.py
+++ b/backend/src/backend/main.py
@@ -82,53 +82,52 @@ async def add_book(book_payload: BookAddSchema):
         zimcheck=book_payload.zimcheck,
     )
 
-    title = await Title.objects.get_or_create(
-        ident=get_ident_from_name(book_payload.metadata["Name"])
-    )
+    for metadata_name, value in book_payload.metadata.items():
+        if metadata_name.startswith("Illustration_"):
+            await BookMetadata.objects.create(
+                book=book.id,
+                name=metadata_name,
+                bin_value=base64.standard_b64decode(value),
+                kind=KIND_ILLUSTRATION,
+            )
+        else:
+            await BookMetadata.objects.create(
+                book=book.id,
+                name=metadata_name,
+                value=value,
+                kind=KIND_TEXT,
+            )
+
+    try:
+        title = await Title.objects.get_or_create(
+            ident=get_ident_from_name(book_payload.metadata["Name"])
+        )
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=f"Exception: {exc}") from exc
+
     await book.update(title=title)
 
-    for metadata_name, value in book_payload.metadata.items():
-        if metadata_name.startswith("Illustration_"):
-            await BookMetadata.objects.create(
-                book=book.id,
-                name=metadata_name,
-                bin_value=base64.standard_b64decode(value),
-                kind=KIND_ILLUSTRATION,
-            )
-        else:
-            await BookMetadata.objects.create(
-                book=book.id,
-                name=metadata_name,
-                value=value,
-                kind=KIND_TEXT,
-            )
-
-    for metadata_name, value in book_payload.metadata.items():
-        if metadata_name.startswith("Illustration_"):
-            await TitleMetadata.objects.create(
-                title=title.ident,
-                name=metadata_name,
-                bin_value=base64.standard_b64decode(value),
-                kind=KIND_ILLUSTRATION,
-            )
-        else:
-            await TitleMetadata.objects.create(
-                title=title.ident,
-                name=metadata_name,
-                value=value,
-                kind=KIND_TEXT,
-            )
+    for metadata in await BookMetadata.objects.all(book=book.id):
+        await TitleMetadata.objects.create(
+            title=title.ident,
+            name=metadata.name,
+            bin_value=metadata.bin_value,
+            value=metadata.value,
+            kind=metadata.kind,
+        )
 
     for tag_name in book_payload.metadata["Tags"].split(";"):
         book_tag = await BookTag.objects.get_or_create(name=tag_name)
         await book.tags.add(book_tag)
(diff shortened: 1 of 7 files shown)


backend/src/backend/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        try:
+            for lang in lang_codes:
+                find_language_names(lang)
+            return f"khan-academy_{'-'.join(lang_codes)}_playlists"
+        except AttributeError:
+            return ""
+
+    else:
+        # this takes care of all the wiki sources
+        lang_codes = name.split("_")[1].split("-")
```
rgaudin, line 22: This could fail is the format is not correct.

backend/src/backend/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        try:
+            for lang in lang_codes:
+                find_language_names(lang)
```
rgaudin, line 25: I don't think `find_language_names` is appropriate to test validity of a language code. Look at the rest of `zimscraperlib.i18n` and if there's nothing quite appropriate, open a ticket on scraperlib repo.

backend/src/backend/main.py
```diff
@@ -68,6 +70,11 @@ async def test(timestamp: int):
 async def add_book(book_payload: BookAddSchema):
     """API endpoint to receive Book addition requests and add to database"""
 
+    ident = get_ident_from_name(book_payload.metadata["Name"])
+    if ident:
```
rgaudin, line 74: I'd prefer this function to fail rather than return an empty string. Ideally, common use cases should return a ValueError with a descriptive message.

backend/src/backend/main.py
```diff
@@ -68,6 +70,11 @@ async def test(timestamp: int):
...
+        # if valid ident
+        title = await Title.objects.get_or_create(ident=ident)
+
```
rgaudin, line 77: We need to fill the Title with data as well. Can be done after the book is created though.

backend/tests/test_stub.py
```diff
@@ -130,3 +126,13 @@ async def test_books_add_endpoint_save_languages(book_dict, base64_png):
...
+
+
+@pytest.mark.asyncio
+async def test_add_book_title(book_dict):
+    response = client.post("/v1/books/add/", json=book_dict)
+    assert response.status_code == 201
+    assert response.headers.get("Content-Type") == "application/json"
+    book = await Book.objects.get(id=book_dict["id"])
+    assert book.title is not None
+    assert book.title.ident == "wikipedia_lg_all"
```
rgaudin, line 138: Once you add the other title data, we will need to test those

rgaudin (changes requested):
Commented inline ; please let me know if anything's not clear.

rgaudin:
Removed Draft status. Draft is for when you're working on it but are not ready for review

backend/tests/conftest.py
```diff
@@ -53,7 +53,41 @@ def book_dict():
             "Tags": "wikipedia;_category:wikipedia;_pictures:yes;"
             "_videos:no;_details:yes;_ftindex:yes",
             "Title": "Wikipedia",
-            "Illustration_48x48": "UklGRv4GAABXRUJQVlA4WAoAAAAQAAAALwAALwAAQUxQSC",
+            "Illustration_48x48": (
```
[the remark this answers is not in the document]
anshulxyz: @rgaudin first I thought I could save visual space by having a small string. Later, after adding more code, it became necessary to have a valid base64 image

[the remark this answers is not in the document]
rgaudin: I see.

backend/src/backend/utils.py
anshulxyz: @rgaudin how should I handle it? 

These are the only book names that didn't work with the method

```
tutorial-wikipedia
wikipedia_roa-tara_all
wikipedia_sh_top
wikistage_multi_all
wikipedia_be-tarask_all
wikipedia_sh_all
wikizimmer.pokepedia-pw_
kiwix.wikipedia_crh-latn_all
wikipedia_eml_all
```

Should I write a regex/pattern, to handle these cases as well?

rgaudin: No, as discussed. First step is getting the 90% that fits. Leave those aside for now.

rgaudin: Your function was assuming it could not fail and would return empty string on error. Users are allowed to send garbage to the endpoint and we have to make sure this doesn't crash and return the appropriate error code/response.

As mentioned elsewhere the failsafe approach is not what we want. In your code calling this function, you'll wrap it to catch exceptions.

Now here you can get an `IndexError` if the text is not composed of multiple `_`. We don't want our user to have to catch IndexError nor is it helpful to the enduser. Instead, you'll wrap this call in a try for `IndexError` and raise a `ValueError` mentionning the fact that the language code(s) could not be found.

anshulxyz: > the failsafe approach is not what we want

I'll keep this in mind

backend/src/backend/utils.py
```diff
@@ -0,0 +1,21 @@
+from zimscraperlib.i18n import NotFound, get_language_details
+
+
+def get_ident_from_name(name: str) -> str:
+    """Get an ident based on the name of the book. Name could be either use mul
+    (yes it's 3 letters). Return pattern should be {project-name}_{lang}_{classifier}
+    """
+    try:
+        lang_codes = name.split("_")[1].split("-")
+    except IndexError as err:
```
rgaudin, line 10: We use `exc` as exception variable by convention (in openzim projects).

backend/src/backend/utils.py
```diff
@@ -0,0 +1,21 @@
+from zimscraperlib.i18n import NotFound, get_language_details
+
+
+def get_ident_from_name(name: str) -> str:
+    """Get an ident based on the name of the book. Name could be either use mul
```
rgaudin, line 5:
```py
    """Title identifier from Book name

     Format: {project}_{langs}_{classifier}
     `langs`: ISO-639-1 lang code or dash-separated list of such or `mul`"""
```

backend/src/backend/utils.py
```diff
@@ -0,0 +1,21 @@
...
+        lang_codes = name.split("_")[1].split("-")
+    except IndexError as err:
+        raise ValueError(
+            f"Can't find language code in '{name}''."
+            " It should follow the pattern of {{project-name}}_{{lang}}_{{classifier}}"
+        ) from err
+
+    for lang in lang_codes:
+        try:
+            get_language_details(lang)
```
rgaudin, line 18: 👍  please open a ticket on scraperlib requestin a direct function like `is_valid_langcode()`.

backend/src/backend/utils.py
```diff
@@ -0,0 +1,21 @@
...
+        except NotFound as err:
+            raise ValueError(f"'{lang}' is not a valid ISO language code.") from err
```
rgaudin, line 20:
```py
            raise ValueError(f"Invalid language code: `{lang}`") from exc
```

backend/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+@pytest.mark.parametrize(
+    "book_name",
+    [
+        "tutorial-wikipedia",
+        "wikipedia",
+    ],
+)
+async def test_extracting_lang(book_name):
+    with pytest.raises(ValueError, match="Can't find language code"):
+        assert get_ident_from_name(book_name)
```
rgaudin, line 56: You don't need the `assert` if you're expecting it to raise

backend/src/backend/main.py
```diff
@@ -95,20 +97,25 @@ async def add_book(book_payload: BookAddSchema):
...
 
+    title = await Title.objects.get_or_create(
+        ident=get_ident_from_name(book_payload.metadata["Name"])
+    )
+    await book.update(title=title)
+
     for tag_name in book_payload.metadata["Tags"].split(";"):
         book_tag = await BookTag.objects.get_or_create(name=tag_name)
         await book.tags.add(book_tag)
+        await title.tags.add(book_tag)
```
rgaudin, line 108: We want only the non-book specific Tags.

Book-specific Tags to exclude:
- `_sw:<no|yes>`
- `_ftindex:<no|yes>`
- `_pictures:<no|yes>`
- `_videos:<no|yes>`
- `_details:<no|yes>`

So it's mostly the *private* tags (starting with an `_`) but the `_category:xxx` one. Please specifically exclude the list above.

rgaudin (changes requested):

rgaudin:
Thank you for this new round ; I've added inline comments.

backend/src/backend/main.py
```diff
@@ -97,25 +103,37 @@ async def add_book(book_payload: BookAddSchema):
...
+                name=metadata_name,
+                value=value,
+                kind=KIND_TEXT,
+            )
 
     for tag_name in book_payload.metadata["Tags"].split(";"):
         book_tag = await BookTag.objects.get_or_create(name=tag_name)
         await book.tags.add(book_tag)
-        await title.tags.add(book_tag)
+        if not tag_name.startswith("_") or "_category" in tag_name:
```
rgaudin, line 125: Oh come on! I **specifically** requested you exclude the list I gave you. It's not even correct as `_category` tag doesn't exist. It's a prefix that we use like `_category:ted` or `_category:wikipedia`.

backend/src/backend/main.py
```diff
@@ -81,6 +82,11 @@ async def add_book(book_payload: BookAddSchema):
         zimcheck=book_payload.zimcheck,
     )
 
+    title = await Title.objects.get_or_create(
```
rgaudin, line 85: You must wrap this in a try to prevent a ValueError from crashing this and return a 400 error instead

backend/src/backend/main.py
```diff
@@ -97,25 +103,37 @@ async def add_book(book_payload: BookAddSchema):
                 kind=KIND_TEXT,
             )
 
-    title = await Title.objects.get_or_create(
-        ident=get_ident_from_name(book_payload.metadata["Name"])
-    )
-    await book.update(title=title)
+    for metadata_name, value in book_payload.metadata.items():
```
rgaudin, line 106: Would make more sense to loop over the BookMetadata to create the BookMetadata. It thus becomes a simple copy and you don't have to do different stuff for illustration.

backend/src/backend/main.py
```diff
@@ -97,25 +103,37 @@ async def add_book(book_payload: BookAddSchema):
...
     for tag_name in book_payload.metadata["Tags"].split(";"):
         book_tag = await BookTag.objects.get_or_create(name=tag_name)
         await book.tags.add(book_tag)
-        await title.tags.add(book_tag)
+        if not tag_name.startswith("_") or "_category" in tag_name:
+            await title.tags.add(book_tag)
 
     for lang_code in book_payload.metadata["Language"].split(","):
-        lang_dict = get_language_details(lang_code)
+        lang_dict = find_language_names(lang_code)
```
rgaudin, line 129: Using the type of a variable in its name is usually not a good idea. It's generally the sign its name if not clear enough. Here it's even wrong as this is a tuple.
Since this function returns a tuple and you'll use both values, create two variables: 

```py
english_name, native_name = find_language_names(lang_code)
```

backend/tests/test_stub.py
```diff
@@ -51,11 +58,17 @@ def test_test_endpoint_invalid_input():
 
 @pytest.mark.asyncio
 async def test_books_add_endpoint(book_dict):
+    await TitleMetadata.objects.delete(each=True)
```
rgaudin, line 61: You should filter on the ident that you'll be creating so that you only delete here what you will create in this test.

By keeping tests isolated, we're saving hard-to-debug issues that could arise as the codebase grow like test unknowingly depending on others and thus randomly failing.

rgaudin (changes requested):
Thanks! There still are some stuff left to fix.

backend/src/backend/main.py
```diff
@@ -95,20 +98,41 @@ async def add_book(book_payload: BookAddSchema):
                 kind=KIND_TEXT,
             )
 
+    try:
+        title = await Title.objects.get_or_create(
+            ident=get_ident_from_name(book_payload.metadata["Name"])
+        )
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=f"Exception: {exc}") from exc
```
rgaudin, line 106: We caught that so we can provide better detail on the failure so inform the caller of the cause:

```py
f"Invalid Name metadata. Unable to build Title identifier: {exc}"
```

backend/src/backend/main.py
```diff
@@ -95,20 +98,41 @@ async def add_book(book_payload: BookAddSchema):
...
+
+    await book.update(title=title)
+
+    for metadata in await BookMetadata.objects.all(book=book.id):
```
rgaudin, line 110: Use `book.metadata.all()` instead. Simpler.

backend/src/backend/main.py
```diff
@@ -95,20 +98,41 @@ async def add_book(book_payload: BookAddSchema):
...
+            name=metadata.name,
+            bin_value=metadata.bin_value,
+            value=metadata.value,
+            kind=metadata.kind,
+        )
+
     for tag_name in book_payload.metadata["Tags"].split(";"):
         book_tag = await BookTag.objects.get_or_create(name=tag_name)
         await book.tags.add(book_tag)
+        if not tag_name.startswith(
```
rgaudin, line 122: Not including the `:` is too loose here. Actually, let's match them properly as suggested

```py
if re.match(r"_(sw|ftindex|pictures|videos|details):(yes|no)", tag_name):
```

backend/tests/test_stub.py
```diff
@@ -227,3 +227,13 @@ async def test_add_book_title_save_metadata(book_dict):
...
+
+
+@pytest.mark.asyncio
+async def test_add_book_title_raise_httpexception_for_title(book_dict):
+    # updating book_dict with incorrect language code
+    book_dict["metadata"]["Name"] = "wikipedia_zz_all"
+    await TitleMetadata.objects.delete(title="wikipedia_lg_all")
+    response = client.post("/v1/books/add/", json=book_dict)
+    assert response.status_code == 400
+    assert response.headers.get("Content-Type") == "application/json"
```
rgaudin, line 239: Great! You should also query the DB to ensure nothing has been created

backend/tests/test_stub.py
```diff
@@ -58,7 +58,7 @@ def test_test_endpoint_invalid_input():
 
 @pytest.mark.asyncio
 async def test_books_add_endpoint(book_dict):
-    await TitleMetadata.objects.delete(each=True)
+    await TitleMetadata.objects.delete(title="wikipedia_lg_all")
```
rgaudin, line 61: Either retrieve this string from the book_dict or create a fixture for this ident so we don't have this static string all over the tests.

rgaudin (changes requested):
Almost there!

Issue #29, opened 2021-10-28
Titles endpoint


rgaudin:
- List titles on `/titles` (includes pagination handling)
- Get title details on `/title/{name}`

Required for #28

anshulxyz:
@rgaudin

> List titles on /titles (includes pagination handling)

Q. What are the properties you want to see in the list? Other than the `ident`?

rgaudin:
Thanks for asking ; it's indeed an important question. Here's how I see the `/titles` REST endpoint:

**Note**: I know this ticket is about the listings only but I'll document it all here.

**GET `/titles`**:

```json
{
  "meta": {
    "count": 2,
    "limit": 10,
    "skip": 0
  },
  "items": [
    {"ident": "wikipedia_fr_all_maxi"},
    {"ident": "wikipedia_fr_all_nopic"}
  ]
}
```

Rule of thumb is that all list endpoints should use the same wrapping schema with the meta info for pagination. It means that it should also accept `skip` and `limit` parameters to control the result list. This is similar to what the [zimfarm API](https://api.farm.openzim.org/v1#/public/listSchedules) does.

Now, this endpoint is gonna be the most used and must be versatile so it serves most if not all of the UI's needs (as well as other needs). It thus needs to allow filtering and payload-control using GET parameters.

**Filters**:
- `?lang=fra,eng` to filter results based on ISO-639-3 lang codes (intersection)
- `?lang=fra|eng` to filter results based on ISO-639-3 lang codes (union)
- `?metadata-name=wikipedia_fr_test` to filter based on a meta (`Name` here). Value can include the `*` character (ex: `wikipedia_*_maxi`, `*_nopic`, etc). `Illustration_*` are not supported obviously.
- `?tag=wikipedia,kids`, `?tag=wikipedia|wikibooks` to filter results based on Tag (intersection and union) with support for exclusion: `tag=wikipedia,!fun|ted,kids` (matches Titles with wikipedia or ted that also have the kids tag and not the fun tag) `!wikipedia` would match all but the wikipedia ones.

**Payload-control**:
- `?with_metadata=title,scraper,illustration_48x48`: includes the requested metadata in response. `all` is a shortcut for including all metadata (including illustration). `all_text` is same but only text-ones. `illustration` is an alias for `illustration_48x48`.
- `?with_tags`: include tags (all) into response.
- `?with_languages`: include languages into response.

[...]

PR #39, opened 2021-12-08
Add `/titles/` and `/titles/{ident}` endpoint 

anshulxyz:
- refer https://github.com/openzim/cms/issues/29#issuecomment-987848283 for details on endpoints

#### - [ ] GET `/titles`

- [x] add pagination
- [ ] add filtering

For pagination I am using https://github.com/uriyyo/fastapi-pagination

Output
```json
{
  "total": 2,
  "page": 1,
  "size": 50,
  "items": [
    {"ident": "wikipedia_fr_all_maxi"},
    {"ident": "wikipedia_fr_all_nopic"}
  ]
}
```

#### - [x] GET `/titles/{ident}`

Output
```json
[
  {
    "ident": "wikihow_fr_all_maxi",
    "languages": [
      "fra"
    ],
    "metadata": {
      "Creator": "Wikipedia",
      "Description": "offline version of Wikipedia in Ganda",
      "Flavour": "maxi",
      "Language": "fra",
      "Name": "wikipedia_fr_all_maxi",
      "Publisher": "Kiwix",
      "Tags": "wikihow;_category:wikipedia",
      "Title": "Wikipedia",
      "Scraper": "wikihow2zim 1.11",
      "Illustration_48x48": "xxx"
    },
    "tags": [
      "wikihow",
      "_category:wikipedia"
    ]
  }
]
```


backend/src/backend/main.py
```diff
@@ -137,4 +137,18 @@ async def add_book(book_payload: BookAddSchema):
     return {"msg": "ok", "uuid": str(book.id), "title": book.title.ident}
 
 
+@database.transaction()
+@api.get("/titles/", status_code=200, response_model=TitlesListSendSchema)
```
rgaudin, line 141: User `/titles` as route path. Will work for both

rgaudin: I also think it's time to split views into different modules.

Let's have something like the following for now:

```
routes/
  __init__.py
  books.py
  titles.py
```

backend/src/backend/main.py
```diff
@@ -137,4 +137,18 @@ async def add_book(book_payload: BookAddSchema):
...
+async def receive_titles():
```
rgaudin, line 142: I don't understand what `receive_titles` mean. Let's use the following convention everywhere:

* `list_titles`
* `get_title`
* `create_title`
* `update_title`
* `delete_title`
* `delete_titles` (we won't have that I believe)

backend/tests/conftest.py
```diff
@@ -11,6 +11,14 @@ def database_url():
     yield
 
 
+@pytest.fixture(autouse=True, scope="function")
```
rgaudin, line 14: We don't want to clean up database after every test. That's bad practice.
We can have explicit calls to clear some or all tables ; or we could have a complete cleanup at the end of the session. Might be better in `database_url` then ; or called there.

backend/src/backend/main.py
```diff
@@ -137,4 +137,18 @@ async def add_book(book_payload: BookAddSchema):
...
+    titles = await Title.objects.select_all().all()
```
rgaudin, line 143: We should explicitly select what we want to return here (`ident`) in order to request just what's required from the DB. Titles are pretty thin at the moment but might evolve.

rgaudin (commented):
Hey, didn't realize at first that it's a draft so I took a quick look and made some early comments.

backend/tests/conftest.py
anshulxyz: hi @rgaudin I have been trying to fix this but couldn't, here's one of the example I tried

```python
@pytest.fixture(scope="function")
async def clear_titles():
    titles = await Title.objects.all()
    for title in titles:
        await title.languages.clear()
        await title.tags.clear()
        await title.metadata.clear()
        await title.delete()
    yield
```

I used this in `test_stub_titles.py`
```python
async def test_titles_endpoint_get_list_single_title(clear_titles, title):
```

but I am getting errors like

```
E               sqlite3.IntegrityError: FOREIGN KEY constraint failed
E               sqlite3.IntegrityError: UNIQUE constraint failed: languages.code
```

anshulxyz:
Hi @rgaudin can you take a look?

backend/tests/conftest.py
rgaudin: yes, this is exactly what I've been warning about. Either we have a clean database for every test (what you've been doing) and tests should be fine but you have no idea what's going on or you clean up everything you created upon teardown of each test. That's what we want. The more we wait to do this, the harder it gets as the number of tests grows.

You sample is good albeit you should yield first and then delete stuff as the part up to the yield statement is called on test setup and the part after is called on test teardown.

I've been able to run all your tests by creating three fixtures: `clear_languages`, `clear_titles`, `clear_books` and `clear_book_dict_languages`. I was just trying to validate that it would work.

Just to be clear, here are the rules we want to follow:

- Model fixtures (such as language for instance) should delete themselves.
- If testing a specific model or a model-specific endpoint (language), it's fine to assume we want to clear the table so a `Model.delete(each=True)` teardown is fine.
- If testing a regular general endpoint (like `books/add`), then we want to remove what we created. That would mean a custom teardown fixture for that endpoint.
- We should not have any removal code at the begining of tests.

Based on this, I believe we'd need the following fixtures:

- `clear_languages()`: `Language.delete(each=True)`
- `clear_books()`: `Language.delete(each=True)` for `test_book_create()`
- `clear_titles()`: for `test_title_create()`
- `clear_book_dict()`: removes metadata, tags, languages, title and book created from the `book_dict`

We may bend those rules in the future but ATM, I see no reason to.

backend/src/backend/routes/titles.py
```diff
@@ -0,0 +1,44 @@
...
+
+    return {
+        "ident": title.ident,
+        "languages": [lang.code for lang in title.languages],
+        "tags": [tag.name for tag in title.tags],
+        "metadata": {
+            (elem.name): (
+                elem.value if elem.value else base64.standard_b64encode(elem.bin_value)
+            )
+            for elem in title_metadata
```
rgaudin, line 42: I don't understand why you are using the `TitleMetadata` model in place of the instance accessor… I've seen that in a test as well I think…

```py
{item.name: (base64.standard_b64encode(item.bin_value) if item.kind == KIND_ILLUSTRATION else item.value)
for item in await title.metadata.all()}
```

backend/src/backend/routes/titles.py
```diff
@@ -0,0 +1,44 @@
...
+    title_metadata = await TitleMetadata.objects.fields("id", "metadata").all(
+        title=title.ident
+    )
+
+    return {
+        "ident": title.ident,
+        "languages": [lang.code for lang in title.languages],
+        "tags": [tag.name for tag in title.tags],
+        "metadata": {
+            (elem.name): (
```
rgaudin, line 39: I find `item` to be more appropriate than `elem` in such case

rgaudin (commented):
Took a look around the PR ; let me know if there are stuff to discuss

backend/tests/conftest.py
```diff
@@ -164,6 +222,56 @@ async def title(title_dict):
 async def title_with_language(title, language_eng):
     await title.languages.add(language_eng)
     yield title
+    await title.languages.clear()
+    await title.tags.clear()
```
rgaudin, line 226: While the title.languages.clear() is fine, the others are unnecessary and redundant with `title` fixture's teardown.

backend/tests/conftest.py
```diff
@@ -164,6 +222,56 @@ async def title(title_dict):
...
+    tag = await TitleTag.objects.create(name="wikipedia")
+    yield tag
+    await tag.delete()
+
+
+@pytest.fixture(scope="function")
+@pytest.mark.asyncio
+async def title_with_data(title, language_eng, title_tag, book_dict):
+    await title.languages.add(language_eng)
+    # await title.tags.add(title_tag)
```
rgaudin, line 243: that should be restored

backend/tests/conftest.py
```diff
@@ -164,6 +222,56 @@ async def title(title_dict):
...
+            await TitleMetadata.objects.create(
+                title=title.ident,
+                name=metadata_name,
+                value=value,
+                kind=KIND_TEXT,
+            )
+    yield title
+    await title.languages.clear()
+    await title.tags.clear()
+    await title.books.clear()
```
rgaudin, line 264: You did not add a book so should not clear it here

rgaudin (changes requested):
Thank you ; we're almost there! A few minor fixes.

I'd also like, as discussed, that you check the DB for undeleted stuff.

Running tests, I see data in `books`, `books_metadata`, `books_tags`, `languages`, an `titles_metadata`. Should be easy to spot now.

backend/tests/conftest.py
anshulxyz: this makes me realize that it is better to have static value in the test, rather than dynamic, because otherwise the tests don't fail because they are checking and asserting from the same data

so I am going to assert like this
```python
    assert response.json() == {
        "ident": title_with_data.ident,
        "languages": ["eng"],
        "tags": ["wikipedia"],
        "metadata": book_dict["metadata"],
    }
```

this way if we don't add tag to the title our test will fail

anshulxyz: this makes me realize that it is better to have static value in the test, rather than dynamic, because otherwise the tests don't fail because they are checking and asserting from the same data

so I am going to assert like this
```python
    assert response.json() == {
        "ident": title_with_data.ident,
        "languages": ["eng"],
        "tags": ["wikipedia"],
        "metadata": book_dict["metadata"],
    }
```

this way if we don't add tag to the title our test will fail

What do you say? @rgaudin

backend/tests/test_titles_endpoint.py
```diff
@@ -25,7 +25,7 @@ async def test_titles_endpoint_get_title(title_with_data, book_dict):
     assert response.status_code == 200
     assert response.json() == {
         "ident": title_with_data.ident,
-        "languages": [lang.code for lang in title_with_data.languages],
-        "tags": [tag.name for tag in title_with_data.tags],
+        "languages": ["eng"],
```
rgaudin, line 28: Why are you doing this? We are using a fixture so we don't want to mix fixture stuff with hard coded values.
Also why wouldn't the tags be present?

rgaudin (changes requested):
Thanks ; please explain/fix the change in test_titles_endpoint and also adress the remaining data in the DB

backend/tests/conftest.py
rgaudin: Ah sorry I missed this comment.

- we don't want to maintain static values at different places. this will be hell to maintain.
- the model-instance fixtures are to provide fake data to the DB so we can query it easily
- we should'nt use those model-instance fixtures as source to assert ; as you mentioned, it's flawed.
- the model-instance fixtures are/should-be built from static values though (like the book_dict). We should use those to compare endpoint-returned data with what we got.

backend/tests/test_titles_endpoint.py
anshulxyz: @rgaudin okay, reverted

anshulxyz:
> and also adress the remaining data in the DB

working on this now

backend/tests/conftest.py
anshulxyz: @rgaudin just saw this, I agree, fixing this now

backend/tests/test_titles_endpoint.py
rgaudin: Thanks but you're right ; as I described in my reply to your comment above ; we should compare with book_dict data (or another static fixture) and not from `title_with_data` itself. Thanks for bringing that up

anshulxyz: @rgaudin to do with I had to update `title_with_data` to have the metadata values same as `book_dict`

rgaudin: 👍

rgaudin (approved):
Thanks!!

PR #40, opened 2021-12-16
UI titles listing and title details display

### LINKED ISSUE #28: Titles listing
UI should present all the titles in the database.

### LINKED ISSUE #30: Title detail
UI should expose all of the Title's data, including the list of associated Books.

anshulxyz:
closes #28 
closes #30 

titles listing

Screenshot 2021-12-16 at 1 45 35 PM

title detail display
Screenshot 2021-12-16 at 2 16 00 PM

Our base64 image data is broken, I tried it with a valid base64 image and it works. For example:

Screenshot 2021-12-16 at 2 21 05 PM


rgaudin (changes requested):
Thank you ; it's good.

I've pushed the linting stuff. I haven't added a rule to pre-commit as pre-commit doesn't support being run from a different directory so that won't work out of the box. We'll work around that later.

Please fix the minor naming stuff I reported inline. In addition:

- remove the testAPI stuff. we don't need it anymore.
- we need error handling. For now, we can, as we do in the zimfarm, display a bootstrap-alert on top of the view with an error message so that it's easy to *implement* error handling in all views.

anshulxyz:
I have added error handling, it's a simpler version of the one we have on Zimfarm.

Screenshot 2021-12-17 at 1 52 52 PM

rgaudin (approved):
Great 👍

PR #42, opened 2021-12-17
Add books' ids to response in /titles/{ident} endpoint

### LINKED ISSUE #31: Book detail
UI should expose Book details either going down Titles or via a direct Book ID search.

Book detail displays all of Book metadata and associated Title.

anshulxyz:
Closes #31

- Backend, add route `GET /books/{id}` to fetch the info of a book entry
- Frontend, add route `/books/{id}` to display the info of the book

#### tasks

- [x] a direct Book ID search
- [x] Book detail displays all of Book metadata and associated Title


anshulxyz:
Screenshot 2021-12-18 at 2 56 47 AM

How do you want to display metadata and zimcheck?

anshulxyz:
Screenshot 2021-12-18 at 2 56 47 AM

@rgaudin  How do you want to display metadata and zimcheck?

backend/src/backend/routes/titles.py
```diff
@@ -40,4 +40,5 @@ async def get_title(ident: str):
             )
             for item in await title.metadata.all()
         },
+        "books": [book.id for book in title.books],
```
rgaudin, line 43: I suggest we return `[{"id": book.id, "name": book.metadata.get(name="Name")}` instead.
This will be useful in any UI-driven context as the Id is not human friendly and would otherwise require requesting the full book endpoint just to get a hint of what's behind the Id.

frontend/src/views/TitleDetail.vue
```diff
@@ -59,6 +59,17 @@
               >
             </td>
           </tr>
+          <tr>
+            <td>Books</td>
+            <td
```
rgaudin, line 64: this won't work as soon as you have more than one book for the title: this table is 2-cols large so it will break the rendering.
You can use a single cell and if you have more than a single book, use a list `ul` inside the cell

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,108 @@
...
+            <td>Article Count</td>
+            <td>{{ book_article_count }}</td>
+          </tr>
+          <tr>
+            <td>Media Count</td>
+            <td>{{ book_media_count }}</td>
+          </tr>
+          <tr>
+            <td>Counter</td>
+            <td>{{ book_counter }}</td>
```
rgaudin, line 20: This is JSON data. Should be formatted in a scrollable `<pre>`

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,108 @@
...
+          </tr>
+          <tr>
+            <td>Metadata</td>
+            <td>{{ book_metadata }}</td>
```
rgaudin, line 24: This should be presented as a table/mapping

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,108 @@
...
+          </tr>
+          <tr>
+            <td>Period</td>
+            <td>{{ book_period }}</td>
+          </tr>
+          <tr>
+            <td>Size</td>
+            <td>{{ book_size }}</td>
```
rgaudin, line 32: Should be humanized. Bytes are not much usable for a user

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,108 @@
...
+          </tr>
+          <tr>
+            <td>Url</td>
+            <td>{{ book_url }}</td>
```
rgaudin, line 36: Should be clickable

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,108 @@
...
+          </tr>
+          <tr>
+            <td>Zimcheck</td>
+            <td>{{ book_zimcheck }}</td>
```
rgaudin, line 40: Again, JSON, so scrollable `pre`

rgaudin (changes requested):
Thank you ; a few comments inline.

Also Issue #31 mentions Searching a Book by its ID so that needs to be added.

backend/tests/test_books_endpoint.py
```diff
@@ -197,3 +196,15 @@ async def test_books_title_raise_httpexception_for_title(book_dict):
...
+
+
+@pytest.mark.asyncio
+async def test_get_book_info(book_with_metadata, book_dict):
+    response = client.get(f"/v1/books/{book_dict['id']}")
+    assert response.status_code == 200
+    assert response.headers.get("Content-Type") == "application/json"
+
+    # removing this key because we don't have this in the receiving schema
+    del book_dict["obsolete"]
```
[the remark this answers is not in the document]
anshulxyz: @rgaudin removing this from the `conftest.py` fixes the issue. In the model we have set it to `False` by default. 

Do we need to remove it from the models too? We'll be adding a migration file.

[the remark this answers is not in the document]
rgaudin: Yes, I though we'd need something like it but it's unclear what we'll need at this stage so let's remove it. It has no value ATM.

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,108 @@
...
+  mixins: [Common],
+  props: {
+    id: {
+      type: String,
+      default: null
+    }
+  },
+  data () {
+    return {
+      book_id: null,
```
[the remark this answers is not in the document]
anshulxyz: @rgaudin I thought we don't want to store the entire response in single variable!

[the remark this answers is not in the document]
rgaudin: No, we don't want to store the `response` which contains the pagination metadata and the list of titles (that's what we discussed before). But once we have our *meaningful object*, we can keep it in a single variable.

So for instance, when retrieving list of books, you can have a `titles` variables with the `data.items` content. When retrieving a book, you can have a `book` variable.

Let me know if that makes sense to you ; I'm not sure I'm expressing it right

backend/src/backend/routes/titles.py
anshulxyz: @rgaudin I have this snippet, let me know if this is okay by you
```py
    books = []
    if title.books:
        for book in title.books:
            name = await book.metadata.get(name="Name")
            books.append({"id": book.id, "name": name.value})
```
because `book.metadata.get(name="Name")` returns a queryObject of type `BookMetadata` and I am getting a `ValidationError` from pyDantic when I do

```
await book.metadata.fields("fields").get(name="Name")
```

rgaudin: Yes, mine was pseudo-code. You should add a `name` property on the book instance that returns the metadata value directly (or None if not present). That seems adequate/useful.

anshulxyz: hi @rgaudin I could not get a one-line solution to work for this, let me know if the current solution works, I have made the commit for you to review

anshulxyz:
> Searching a Book by its ID so that needs to be added

working on this now, let me know if the following approach is okay

1. Frontend - a new page, with a search bar, to send the book id to the backend
2. Backend - `GET /books/{id}` endpoint receives the id, and sends back the details, if book found, else 404

anshulxyz:
@rgaudin 
> Searching a Book by its ID so that needs to be added

working on this now, let me know if the following approach is okay

1. Frontend - a new page, with a search bar, to send the book id to the backend
2. Backend - `GET /books/{id}` endpoint receives the id, and sends back the details, if book found, else 404

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,126 @@
...
+              <table>
+                <tbody>
+                  <tr
+                    v-for="(value, key) in book.metadata"
+                    :key="value"
+                  >
+                    <td>
+                      {{ key }}
+                    </td>
+                    <td v-if="key === 'Illustration_48x48'">
```
rgaudin, line 50: Should check for the kind ; not this very specific name as ZIM can have multiple illustratiobs

frontend/src/views/BookDetail.vue
```diff
@@ -0,0 +1,126 @@
...
+  mixins: [Common],
+  props: {
+    id: {
+      type: String,
+      default: null
+    }
+  },
+  data () {
+    return {
+      accountBalance: 200,
```
rgaudin, line 103: ??

rgaudin (changes requested):

frontend/src/views/BookDetail.vue
anshulxyz: Hi @rgaudin  I am using the `startsWith('Illustration_')`

because in order to add "kind" data to the response I would have to change a lot of structure

and we use `startswith()` for the same work in Python

anshulxyz:
@rgaudin here's a demo of the search

https://user-images.githubusercontent.com/6121530/147331786-1abca405-b946-4393-a4af-c9b1e9ca9ce8.mov

frontend/src/views/BookDetail.vue
rgaudin: OK that's good enough for now

rgaudin (approved):
Thank you

PR #45, opened 2021-12-29
Create UI testing fixtures

anshulxyz:
Add data fixtures for testing UI

Command
```
invoke db-upgrade load-demo-fixture
```

You need to have migrations before you can load the fixtures.

Or try to make a clean start with
```
rm src/dev.db && invoke db-upgrade load-demo-fixture
```


backend/src/backend/models.py
```diff
@@ -148,7 +148,11 @@ class Meta(BaseMeta):
     title: Optional[Title] = ormar.ForeignKey(Title, related_name="books")
 
     async def book_name(self) -> str:
-        return (await self.metadata.get(name="Name")).value
+        try:
+            return (await self.metadata.get(name="Name")).value
+        except ormar.exceptions.NoMatch:
+            # if the book doesn't have metadata Name
+            return ""
```
rgaudin, line 155: Great but we should return `None`

backend/tests/test_models.py
```diff
@@ -49,6 +49,7 @@ async def test_book_create(book_dict, clear_book_dict):
     assert book.article_count == book_dict["article_count"]
     assert book.size == book_dict["size"]
     assert book.zimcheck == book_dict["zimcheck"]
+    assert await book.book_name() == ""
```
rgaudin, line 52: To fix according to change above

backend/tests/test_models.py
```diff
@@ -115,12 +116,13 @@ async def test_book_illustrations(book, base64_png):
...
 ######
 @pytest.mark.asyncio
-async def test_title_create(clear_titles):
+async def test_title_create():
     ident = "wikipedia_bm_test"
     title = await Title.objects.create(
         ident=ident,
     )
     assert title.ident == ident
+    await title.delete()
```
rgaudin, line 125: Why are you doing this? Using a fixture, we are sure it is going to be executed, regardless of the passing or not of the test. If an exception arise, the rest of the test is not executed and your deletion code is never called.

backend/tests/test_models.py
```diff
@@ -142,13 +144,16 @@ async def test_title_languages(title_with_language, language_fra):
...
-    tag = await TitleTag.objects.create(name="_sw:yes")
-    assert tag.private
+async def test_title_tag_create():
+    tag_1 = await TitleTag.objects.create(name="wikipedia")
+    assert not tag_1.private
+    tag_2 = await TitleTag.objects.create(name="_sw:yes")
+    assert tag_2.private
     assert await TitleTag.objects.count() == 2
 
+    await tag_1.delete()
```
rgaudin, line 154: Same comment as above

rgaudin (changes requested):
Great! Please fix those minor stuff inline

backend/src/backend/models.py
anshulxyz: @rgaudin we're using following schema for this
```py
class BooksListSendSchema(BaseModel):
    id: uuid.UUID
    name: str
```
If I return `None` it will fail the validation of `name: str`

rgaudin: A book cannot not have a Name. So the proper behavior is probably to let the ormar exception raise (or raise a custom InvalidBook Exception).
Haven't checked but I believe our add book endpoint prevents creating Book without a Name. If not, we should.

anshulxyz: @rgaudin 
> I believe our add book endpoint prevents creating Book without a Name. If not, we should.
In the schema of receiving Book data, we have the `metadata` dict as mandatory
```py
class BookAddSchema(BaseModel):
    id: uuid.UUID
    period: datetime.date
    counter: Dict[str, Any]
    article_count: int
    media_count: int
    size: int
    metadata: Dict[str, Any]
    url: pydantic.HttpUrl
    zimcheck: Dict[str, Any]
```
But this doesn't check if the `Name` key is present in it.
We're able to create Book without that. We should probably add another layer of validation?

rgaudin: Yep, I'll create an issue

rgaudin (approved):
👍

PR #49, opened 2022-01-13
Add query filtering parameters for `/titles`

### LINKED ISSUE #41: Add filtering to the `/titles` listing endpoint
refer https://github.com/openzim/cms/issues/29#issuecomment-987848283

**GET `/titles`**:

```json
{
  "meta": {
    "count": 2,
    "limit": 10,
    "skip": 0
  },
  "items": [
    {"ident": "wikipedia_fr_all_maxi"},
    {"ident": "wikipedia_fr_all_nopic"}
  ]
}
```

Rule of thumb is that all list endpoints should use the same wrapping schema with the meta info for pagination. It means that it should also accept `skip` and `limit` parameters to control the result list. This is similar to what the [zimfarm API](https://api.farm.openzim.org/v1#/public/listSchedules) does.

Now, this endpoint is gonna be the most used and must be versatile so it serves most if not all of the UI's needs (as well as other needs). It thus needs to allow filtering and payload-control using GET parameters.

**Filters**:
- `?lang=fra,eng` to filter results based on ISO-639-3 lang codes (intersection)
- `?lang=fra|eng` to filter results based on ISO-639-3 lang codes (union)
- `?metadata-name=wikipedia_fr_test` to filter based on a meta (`Name` here). Value can include the `*` character (ex: `wikipedia_*_maxi`, `*_nopic`, etc). `Illustration_*` are not supported obviously.
- `?tag=wikipedia,kids`, `?tag=wikipedia|wikibooks` to filter results based on Tag (intersection and union) with support for exclusion: `tag=wikipedia,!fun|ted,kids` (matches Titles with wikipedia or ted that also have the kids tag and not the fun tag) `!wikipedia` would match all but the wikipedia ones.

**Payload-control**:
- `?with_metadata=title,scraper,illustration_48x48`: includes the requested metadata in response. `all` is a shortcut for including all metadata (including illustration). `all_text` is same but only text-ones. `illustration` is an alias for `illustration_48x48`.
- `?with_tags`: include tags (all) into response.
- `?with_languages`: include languages into response.

Complete complex request: `GET /titles?lang=fra&metadata-scraper=*1.11&tag=!wikipedia&with_tags&with_languages&with_medata=all`

```json
[
  {
    "ident": "wikihow_fr_all_maxi",
    "languages": [
      "fra"
    ],
    "metadata": {
      "Creator": "Wikipedia",
      "Description": "offline version of Wikipedia in Ganda",
      "Flavour": "maxi",
      "Language": "fra",
      "Name": "wikipedia_fr_all_maxi",
      "Publisher": "Kiwix",
      "Tags": "wikihow;_category:wikipedia",
      "Title": "Wikipedia",
      "Scraper": "wikihow2zim 1.11",
      "Illustration_48x48": "xxx"
    },
    "tags": [
      "wikihow",
      "_category:wikipedia"
    ]
  }
]
```

anshulxyz:
Closes #41 

This is a work-in-progress.

@rgaudin 

I have failing BUT valid tests in this PR.

We will use/need `exclude_unset` refer this https://pydantic-docs.helpmanual.io/usage/exporting_models/
 whether fields which were not explicitly set when creating the model should be excluded from the returned dictionary; default False

I am still confused how they are using it here in FastApi https://fastapi.tiangolo.com/tutorial/body-updates/#using-pydantics-exclude_unset-parameter

here's ORMar's same docs for unsetting https://collerek.github.io/ormar/models/methods/#exclude_unset

---

following is my notes on this issue

-  [x] https://github.com/openzim/cms/issues/41
	- filters
		- [ ]  `?metadata-name=wikipedia_fr_test` to filter based on a meta (`Name` here). Value can include the `*` character (ex: `wikipedia_*_maxi`, `*_nopic`, etc). `Illustration_*` are not supported obviously.
			- [ ] exact match, `?metadata-name=wikipedia_fr_test`
			- [ ] wildcard match, `?metadata-name=wikipedia_*_all`
				- look into **SELECT LIKE**
			- [ ] ignore `Illustration_*`
		- [ ] `?tag=wikipedia,kids`, `?tag=wikipedia|wikibooks`
- [ ] payload-control
	- [ ] `?with_metadata=title,scraper,illustration_48x48`: includes the requested metadata in response. `all` is a shortcut for including all metadata (including illustration). `all_text` is same but only text-ones. `illustration` is an alias for `illustration_48x48`.
	- [ ] **`?with_languages`: include languages into response**
		- the `?with_languages` just by itself is not supported, it needs some value
			- ```diff
				-?with_languages
				+?with_languages=true
		- test
		- add `False` as the default
		- use the `bool` type
		- ❌ the `fields()` is not having any effect
		- test cases
			- [x] titles with single language
			- [x] titles with multiple languages
		- ⭕ Should all the `language` response follow the above style, instead of `["eng", "fra"]`
	- [ ] `?with_tags`: include tags (all) into response.
		- test cases
			- [x] combine `with_languages` and `with_tags`

**Meeting with Renaud** at 17:30
- Q. Will we ever have Titles w/o languages?
	- No
- Q. Should we be sending language data of titles by default. Because `fields("ident")` is not able to remove languages when `select_related()` is used
	- unless `with_languages` is used, don't send "language" key in the response
- Q. Will there be test case where `?lang=fra&with_languages=false`?
	- Yes

**Notes**
- We could just not use pydantic `response_model` for parsing, and then we can customise the response. Note of the pagination working with this new custom response. Refer `/titles/{ident}` on how to do this.
- Code is becoming spaghetti when I try to get multiple query parameters to work together, a.k.a. complex request
- Maybe if we use the difference between `None` vs `[]`, use can still use Pydantic schema?

### DIFF
--- a/backend/src/backend/constants.py
+++ b/backend/src/backend/constants.py
@@ -1,10 +1,13 @@
 import os
+import pathlib
 from dataclasses import dataclass
+
+src_dir = pathlib.Path(__file__).parent.resolve()
 
 
 @dataclass
 class BackendConf:
-    database_url: str = os.getenv("DATABASE_URL", "sqlite:///dev.db")
+    database_url: str = os.getenv("DATABASE_URL", f"sqlite:////{src_dir}/dev.db")
     allowed_origins = os.getenv(
         "ALLOWED_ORIGINS",
         "http://localhost|http://localhost:8000|http://localhost:8080",  # dev fallback
--- a/backend/src/backend/main.py
+++ b/backend/src/backend/main.py
@@ -24,7 +24,7 @@ async def shutdown() -> None:
         await database.disconnect()
 
 
-def create_app():
+def create_app() -> FastAPI:
     app = FastAPI(
         title=__title__,
         description=__description__,
--- a/backend/src/backend/models.py
+++ b/backend/src/backend/models.py
@@ -206,3 +206,51 @@ class TitleMetadata(ormar.Model, MetadataMixin):
         constraints = [ormar.UniqueColumns("name", "title")]
 
     title: Title = ormar.ForeignKey(Title, related_name="metadata")
+
+async def get_matched_m2m_combination(on: str, items: List[str]) -> List[str]:
+    """Ids of requested model matching a combination of values on many2many column
+
+    params:
+        on: a {model}-{collection} string to indiquate which model and m2m colums
+        are being queried.
+        model is either `title` or `book`; collection is either `language` or `tag`
+
+        items: list of pk from the m2m target that must all be present to match
+
+    Sample uses:
+        get list of titles ident for which the m2m has eng and french language:
+            on="title-language", items=["eng", "fra"]
+        get list of books uuids for which the m2m has "wiki", "new" and "kid" tags:
+            on="book-tag", items=["wiki", "new", "kid"]"""
+    model_name, collection_name = on.lower().split("-", 1)
+    if model_name not in ("title", "book") or collection_name not in (
+        "language",
+        "tags",
+    ):
+        raise ValueError("Invalid m2m combination request")
+    if model_name == "title":
+        model = Title
+        id_field = "ident"
+        lookup_field = "title"
+    elif on == "book":
+        model = Book
+        id_field = "uuid"
+        lookup_field = "book"
+    coll_lookup_field = collection_name
+    # fine as long as there's no customed-table-name for the through model
+    through_tablename = f"{model.Meta.tablename}_{collection_name}s"
+
+    # expanding bound params alternative (https://stackoverflow.com/a/25168069/631925)
+    params = {f"item{idx}": item for idx, item in enumerate(items)}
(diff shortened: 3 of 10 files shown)


anshulxyz:
Hi @rgaudin all the tests are passing now and I have added a docs. This docstring comment also shows up on the docs page. At `v1/docs#/titles/list_titles_titles_get`

> WARNING: When the flags are set to false (by default), the API still sends
    those fields in response. The value of those keys are either empty of None.

anshulxyz:
More tests needs to be written, for the test coverage and some more cases.

rgaudin (changes requested):
@anshulxyz, this implementation doesn't comply with the requirements: it is not possible to combine queries. The various query operators (lang, tags, metadata) should all be chained to build up the result so one can request `/titles/?lang=fra|eng&with_metadata=Title&with_languages=y&with_tags=y`

Additionally, the shortcut for all metadata or all_text are not implemented neither.

Additionally., the per-meta filter is not implemented and the search-string neither

PR #54, opened 2022-02-03
Update the Book size from Integer to BigInteger

### LINKED ISSUE #53: Incorrect column type for size
Book column `size` is an `sqlalchemy.Integer` which doesn't fit our bytes ZIM sizes. It should probably be changed to `BigInteger`.

`size=10316940597` results in:

```
pymysql.err.DataError: (1264, "Out of range value for column 'size' at row 1")
```

anshulxyz:
Closes #53 

@rgaudin I ran the migration but the migration files was empty.


rgaudin:
@anshulxyz, I've fixed the alembix conf that I broke when introducing mysql support.
Also added config so column type changes are checked in autogen ([not default behavior](https://alembic.sqlalchemy.org/en/latest/autogenerate.html)) and I eventually added a third party that checks for difference between schema and models so that you can call db-gen in a similar way as Django.

Please check that we haven't missed any migrations and include the one for this

anshulxyz:
Added the migration file for the change in datatype from int to BigInt for book's Size.

> Please check that we haven't missed any migrations and include the one for this

@rgaudin I checked. No other change was made to `models.py`, that needs migration.

rgaudin (approved):
👍

PR #55, opened 2022-02-08
Update migration setting since SQLite doesn't support ALTERing Columns

anshulxyz:
This fixes the following error

> sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "ALTER": syntax error

More info here:
- https://github.com/miguelgrinberg/Flask-Migrate/issues/252
- https://alembic.sqlalchemy.org/en/latest/batch.html#batch-mode-with-autogenerate
- https://stackoverflow.com/a/32510603/3832086


rgaudin (approved):
Thanks

PR #57, opened 2022-02-15
Update demo data Langauage lang-code to ISO-639-3

### LINKED ISSUE #56: Demo data uses incorrect language codes
Although we haven't enforced it at database level (we should check zimfarm for lack of exception), we require the Language codes to be ISO-639-3 which is a 3 letter code.

Not using it in the demo fixture leads to confusion.

anshulxyz:
Closes #56

### DIFF
--- a/backend/src/demo/data.py
+++ b/backend/src/demo/data.py
@@ -3,7 +3,7 @@ import base64
 import re
 import uuid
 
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
Great, thanks

PR #61, opened 2022-03-17
Title reconciliation script

### LINKED ISSUE #38: Add book-title reconciliation script
This is a follow up issues for #25 
also refer PR #37 for where to use this script

Title pattern to follow
```
{project-name}_{lang}_{classifier}
```

fadiga:
Fixes #38


rgaudin:
There is no detail about how the script should be run nor what should be expected from running it.

Running the script, I have a lot of lines like `Name is not remove list.remove(x): x not in list`. Is this expected? What does this mean?

Please be more verbose in the names of the *Table of stats* as understanding exactly what each number mean is important.

I see the CSV files and those look useful. Two comments:
- what does `zimfarm names not in library` mean?
- library_zim_names includes invalid ones. The *aimhi* ones are all wrong. I think we have a regexp for valid names in the ticket but basically, Spaces or parenthesis are not allowed.

👍

kelson42:
I guess you need this PCRE regexp https://github.com/kiwix/maintenance/blob/master/library-docker/bin/manageContentRepository.pl#L109

Issue #50, opened 2022-02-02
Transaction not being rolled back


rgaudin:
When creating a book using `create_book` endpoint, should an exception occur in the metadata handling, the function is halted but the `Book` entry, which is created first, is still present.

The `@database.transaction()` is thus not doing what we'd expect: rollback every query from that function should an exception be raised.

This happens on online version which is using mariadb upon `pymysql.err.IntegrityError: (1062, "Duplicate entry 'Creator-wikipedia_fa_top' for key 'uc_titles_metadata_name_title'")`

PR #63, opened 2022-04-05
Fixed transaction bug and added his test

fadiga:
fixe #50


rgaudin:
Good! Please add a brief explanation in the commit comment as it wasn't obvious

PR #64, opened 2022-04-07
Fixed XML Library Digester.

### LINKED ISSUE #32: Library XML trigger
UI should offer a link to trigger and retrieve (download) the XML output of the Library XML digester

fadiga:
Closes #32


backend/src/backend/digesters/raw.py
```diff
@@ -0,0 +1,56 @@
...
+from backend.models import KIND_ILLUSTRATION, Book
+
+
+async def gen_raw_digester() -> bytes:
+
+    lib_elem = Element("Library")
+    lib_elem.attrib["version"] = "20110515"
+    for book in await Book.objects.select_related("metadata").exclude(title=None).all():
+        book_dict = {}
+        book_dict.update(
```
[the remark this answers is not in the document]
rgaudin: Also, the name is not descriptive of what's inside.

backend/src/backend/digesters/raw.py
```diff
@@ -0,0 +1,56 @@
...
+                item.name: (
+                    base64.standard_b64encode(item.bin_value)
+                    if item.kind == KIND_ILLUSTRATION
+                    else item.value
+                )
+                for item in book.metadata
+            }
+        )
+
+        metadata_name_dic = {
```
rgaudin, line 25: Again, the name is not appropriate

backend/src/backend/digesters/raw.py
```diff
@@ -0,0 +1,56 @@
+import base64
+from urllib.parse import urlparse
+from xml.etree.ElementTree import Element, SubElement, tostring
+
+from backend.models import KIND_ILLUSTRATION, Book
+
+
+async def gen_raw_digester() -> bytes:
+
+    lib_elem = Element("Library")
```
rgaudin, line 10: I think it's not necessary to use etree for the whole library:
- it will require keeping a full, growing copy of the tree in memory while you iterate and then double it to convert to string
- it's probably not efficient on a super large list.
- it's not mandatory as what's outside the `<book />` nodes is ridiculous.

We should thus keep it for book and just fill-up an xml string

backend/tests/test_digesters.py
```diff
@@ -0,0 +1,46 @@
...
+from backend.digesters.raw import gen_raw_digester
+from backend.models import Book
+
+
+@pytest.mark.asyncio
+async def test_raw(client):
+    response = await client.get("/v1/digesters/raw")
+    assert response.status_code == 200
+    assert response.headers.get("Content-Type") == "application/xml"
+    assert response.content == b'<Library version="20110515" />'
```
rgaudin, line 13: We should test the endpoint with data as well

rgaudin (changes requested):
Please see inline comments. Also, this **does not** closes #32 at all

rgaudin (approved):
👍

PR #81, opened 2022-07-01
Actual implementation of Kiwix Public Collection

### LINKED ISSUE #75: Kiwix Public Library Collection
We want to introduce the concept of *Collection* which is a list of Book entries.

This will be part code, part manual selection, ultimately, allowing manual exclusion and insertion. At the moment, we'll solely rely on code, that would mostly be some kind of filter that produce a list (technically a query or generator) of Book entries that can be piped into an *Exporter*.

fadiga:
close #75

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -420,3 +420,27 @@ async def titles_with_metadata(language_fra, language_eng):
         for metadata in await title.metadata.all():
             await metadata.delete()
         await title.delete()
+
+
+@pytest.fixture(scope="function")
+@pytest.mark.asyncio
+async def titles_with_metadata_books(client, book_dict):
+
+    ids = []
+    book_dict["metadata"]["Name"] = "wikipedia_en_all"
+    book_dict["metadata"]["Flavour"] = "maxi"
+    book_dict["metadata"]["Language"] = "eng"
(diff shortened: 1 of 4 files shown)


rgaudin (changes requested):
Thank you; it looks mostly fine.

I'm concerned about the huge amount of duplicated model-manipulation code in `titles_with_metadata_books()`. This looks complicated, errorprone and hard to maintain.
With a fixture that call add_book, that would be easier I believe

backend/tests/conftest.py
```diff
@@ -420,3 +420,90 @@ async def titles_with_metadata(language_fra, language_eng):
...
             await metadata.delete()
         await title.delete()
+
+
+@pytest.fixture(scope="function")
+@pytest.mark.asyncio
+async def titles_with_metadata_books(language_fra, book_dict):
+    """Title with a combination of books with and without Flavours"""
+
+    title_en = await Title.objects.create(ident="wikipedia_en_maxi")
```
rgaudin, line 430: Following the [Convention](https://github.com/openzim/overview/wiki/Naming-Convention), can you use `wikipedia_en_all_maxi` ?

rgaudin (approved):

PR #84, opened 2022-07-11
Auto deleting books

### LINKED ISSUE #71: Auto deleting books
Books are created via a Zimfarm call.

At that moment, ZIMs are probably online or about to be.
We add them to the Title (or create it if it's the first) and that's it.

Over time, we end end with many books for a Title.

We know that we are currently keeping at most 2 dates ZIM files per content on the server so if we have 3 books for a Title in the DB, chances are one of them is not reachable.

We should write down two things IMO:

- what do we do now about this? I suggest we run a periodic script that checks titles with more than 2 books, check if the file is still online (HEAD) and delete the Book if not. We may want to check all the books to unsure we don't expose a dead book but that seems overkill until we use the generated XML file.
- describe what we would ideally want. We've discussed we'd like the CMS to trigger the deletion of expired files but we haven't defined what an expired file is. I suppose this might depend on inclusion of not in one of the libraries…

fadiga:
close #71

### DIFF
--- a/backend/src/backend/constants.py
+++ b/backend/src/backend/constants.py
@@ -1,12 +1,22 @@
+import logging
 import os
 import pathlib
 from dataclasses import dataclass
 
+from zimscraperlib.logging import getLogger as lib_getLogger
+
 src_dir = pathlib.Path(__file__).parent.resolve()
+
+logger = lib_getLogger(
+    src_dir.name,
+    level=logging.INFO,
+    log_format="[%(asctime)s] %(levelname)s:%(message)s",
+)
 
 
 @dataclass
 class BackendConf:
+    nb_book_versions_to_keep = os.getenv("NB_BOOK_VERSIONS_TO_KEEP", 2)
     database_url: str = os.getenv("DATABASE_URL", f"sqlite:////{src_dir}/dev.db")
     allowed_origins = os.getenv(
         "ALLOWED_ORIGINS",
--- a/backend/src/backend/routes/books.py
+++ b/backend/src/backend/routes/books.py
@@ -6,6 +6,8 @@ from fastapi.responses import JSONResponse
 from ormar.exceptions import NoMatch
 from zimscraperlib.i18n import find_language_names
 
+from backend import utils
+from backend.constants import logger
 from backend.models import (
     BOOK_ONLY_METADATA,
     KIND_ILLUSTRATION,
@@ -109,6 +111,15 @@ async def create_book(book_payload: BookAddSchema):
         )
         await book.languages.add(language)
         await title.languages.add(language)
+    flavour = book_payload.metadata.get("Flavour", "")
+    try:
+        # use via utils.xxx to accomodate monkeypatch in tests
+        await utils.remove_obsolete_books(title_ident=title.ident, flavour=flavour)
+    except Exception as exc:
+        logger.warning(
+            f"Unable to remove obsolete books for {title.ident}/{flavour}: {exc}"
+        )
+        logger.exception(exc)
 
     return {"uuid": str(book.id), "title": book.title.ident}
 
--- a/backend/src/backend/utils.py
+++ b/backend/src/backend/utils.py
@@ -1,5 +1,8 @@
 from zimscraperlib.i18n import NotFound, get_language_details
 
+from backend.constants import BackendConf
+from backend.models import Book
+
 
 def get_ident_from_name(name: str) -> str:
     """Title identifier from Book name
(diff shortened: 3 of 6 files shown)


backend/src/backend/utils.py
```diff
@@ -18,3 +21,23 @@ def get_ident_from_name(name: str) -> str:
         except NotFound as exc:
             raise ValueError(f"Invalid language code: `{lang}`") from exc
     return name
+
+
+async def clean_old_books(title_ident: str, flavour: str):
```
rgaudin, line 26: `remove_obsolete_books` would be more accurate

backend/src/backend/utils.py
```diff
@@ -18,3 +21,23 @@ def get_ident_from_name(name: str) -> str:
...
+    """Clean the old delivered so that there are only two recent versions remain"""
```
rgaudin, line 27: `""" Remove matching books for which there are N newer versions available """`

backend/src/backend/utils.py
```diff
@@ -18,3 +21,23 @@ def get_ident_from_name(name: str) -> str:
...
+
+    keep_book_list = []
+    for book in (
+        await Book.objects.exclude_fields(["counter", "languages", "zimcheck", "tags"])
+        .filter(title__ident=title_ident)
+        .select_related("metadata")
+        .order_by("-period")
+        .all()
+    ):
+        if keep_book_list.count(flavour) < BackendConf.number_books_keep:
```
rgaudin, line 37: This should be `<=` otherwise you'll only keep a single book (while wanting 2)

backend/src/backend/constants.py
```diff
@@ -7,6 +7,7 @@
 
 @dataclass
 class BackendConf:
+    number_books_keep = 2
```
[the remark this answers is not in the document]
rgaudin: `nb_book_versions_to_keep` is a more appropriate name, `NB_BOOK_VERSIONS_TO_KEEP` for the envrion.

backend/src/backend/routes/books.py
```diff
@@ -110,6 +110,13 @@ async def create_book(book_payload: BookAddSchema):
         await book.languages.add(language)
         await title.languages.add(language)
 
+    await clean_old_books(
+        title_ident=title.ident,
+        flavour=""
+        if not await book.metadata.filter(name="Flavour").exists()
```
rgaudin, line 116: Don't make a DB request to get this! You already have it available.

backend/src/backend/routes/books.py
```diff
@@ -110,6 +110,13 @@ async def create_book(book_payload: BookAddSchema):
         await book.languages.add(language)
         await title.languages.add(language)
 
+    await clean_old_books(
```
rgaudin, line 113: I think this should be wrapped in a try block as this is a subsidiary, internal operation that we trigger upon new book creation. We wouldn't want the book addition to fail (and thus everything rolled-back) just because the obsolete book deletion failed. This should just trigger a log entry.

rgaudin (changes requested):
Thank you ; please see inline comments

backend/src/backend/utils.py
```diff
@@ -18,3 +21,23 @@ def get_ident_from_name(name: str) -> str:
         except NotFound as exc:
             raise ValueError(f"Invalid language code: `{lang}`") from exc
     return name
+
+
+async def clean_old_books(title_ident: str, flavour: str):
+    """Clean the old delivered so that there are only two recent versions remain"""
+
+    keep_book_list = []
```
rgaudin, line 29: Poor naming

backend/src/backend/constants.py
```diff
@@ -1,12 +1,24 @@
...
+    level=logging.INFO,
+    log_format="[%(asctime)s] %(levelname)s:%(message)s",
+)
+
 
 @dataclass
 class BackendConf:
+
+    nb_book_versions_to_keep = os.getenv("NB_BOOK_VERSIONS_TO_KEEP", 2)
+
```
rgaudin, line 21: why is it surrounded by empty lines?

backend/src/backend/routes/books.py
```diff
@@ -109,6 +111,14 @@ async def create_book(book_payload: BookAddSchema):
...
         await book.languages.add(language)
         await title.languages.add(language)
+    try:
+        # use via utils.xxx to accomodate monkeypatch in tests
+        await utils.remove_obsolete_books(
+            title_ident=title.ident,
+            flavour=book_payload.metadata.get("Flavour", ""),
+        )
+    except Exception as exc:
+        logger.exception(exc)
```
rgaudin, line 121: Please add an error log entry in addition to the exception trace so it's easier to find

rgaudin (changes requested):
See inline

rgaudin (approved):
OK ; let's rebase

PR #90, opened 2022-09-03
Added exporter

### LINKED ISSUE #77: Exporter mechanism and Public Kiwix XML Export
With our Kiwix Library Collection and its Kiwix XML Exporter, we need to actually export that XML into a file on the production server.

How this initial exporter will be implemented is TDB but options are:
- A push of the content to a (non-existing atm) service on the destination server
- A simple call to a (non-existing atm) service on the destination server that will then make a web request to the CMS to fetch data
- An scp/sftp command to a (not setup atm) SSH service on the destination server

This feature might also be used in the future to upload other collections to S3 for instance.

fadiga:
Fixes #77

### DIFF
--- a/backend/src/backend/exporters/__init__.py
+++ b/backend/src/backend/exporters/__init__.py
@@ -1,3 +1,3 @@
 class ExporterInterface:
-    async def exporter(self):
+    async def export(self):
         raise NotImplementedError()
--- a/backend/src/backend/exporters/__init__.py
+++ b/backend/src/backend/exporters/__init__.py
@@ -1,20 +1,3 @@
-import pathlib
-import tempfile
-
-import openzim_uploader
-
-from backend.constants import BackendConf
-
-
-async def exporter(data):
-    src_path = pathlib.Path(tempfile.NamedTemporaryFile(delete=False).name)
-    with open(src_path, "w") as f:
-        f.write(data)
-        f.close()
-
-    await openzim_uploader.check_and_upload_file(
-        src_path=src_path,
-        upload_uri=BackendConf.upload_uri,
-        private_key=BackendConf.private_key,
-        delete=True,
-    )
+class ExporterInterface:
+    async def exporter(self):
+        raise NotImplementedError()
--- a/backend/src/backend/exporters/__init__.py
+++ b/backend/src/backend/exporters/__init__.py
@@ -1,0 +1,20 @@
+import pathlib
+import tempfile
+
+import openzim_uploader
+
+from backend.constants import BackendConf
+
+
+async def exporter(data):
+    src_path = pathlib.Path(tempfile.NamedTemporaryFile(delete=False).name)
+    with open(src_path, "w") as f:
+        f.write(data)
+        f.close()
+
+    await openzim_uploader.check_and_upload_file(
+        src_path=src_path,
+        upload_uri=BackendConf.upload_uri,
+        private_key=BackendConf.private_key,
+        delete=True,
+    )
--- a/backend/src/backend/exporters/kiwix_public_exporter.py
+++ b/backend/src/backend/exporters/kiwix_public_exporter.py
@@ -1,0 +1,10 @@
+from backend.collections.kiwix_public import KiwixPublicCollection
+from backend.exporters import exporter
+from backend.formatters.kiwixlibraryxml import KiwixLibraryXml
+
+
+class KiwixPublicExporter:
+    async def kiwix_library_xml_exporter():
+        await exporter(
+            await KiwixLibraryXml.generate(collection=KiwixPublicCollection())
+        )
--- a/backend/src/backend/exporters/kiwix_public_exporter.py
+++ b/backend/src/backend/exporters/kiwix_public_exporter.py
@@ -15,7 +15,7 @@ class KiwixPublicExporter(ExporterInterface):
         with open(src_path, "w") as f:
             f.write(await KiwixLibraryXml.generate(collection=KiwixPublicCollection()))
 
-        await openzim_uploader.check_and_upload_file(
+        openzim_uploader.check_and_upload_file(
             src_path=src_path,
             upload_uri=BackendConf.upload_uri,
             private_key=BackendConf.private_key,
--- a/backend/src/backend/exporters/kiwix_public_exporter.py
+++ b/backend/src/backend/exporters/kiwix_public_exporter.py
@@ -10,27 +10,26 @@ from backend.formatters.kiwixlibraryxml import KiwixLibraryXml
 
 
 class KiwixPublicExporter(ExporterInterface):
-    """
-    Export kiwix library collection in kiwix xml and upload on
-    the production server
-    """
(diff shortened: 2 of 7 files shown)


kelson42:
At which stage the quality assessment of this new feed will be checked (against current one)?

backend/src/backend/exporters/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+import openzim_uploader
+
+from backend.constants import BackendConf
+
+
+async def exporter(data):
+    src_path = pathlib.Path(tempfile.NamedTemporaryFile(delete=False).name)
+    with open(src_path, "w") as f:
+        f.write(data)
+        f.close()
```
rgaudin, line 13: No need to close when using the context manager

backend/src/backend/exporters/__init__.py
```diff
@@ -0,0 +1,20 @@
+import pathlib
+import tempfile
+
+import openzim_uploader
+
+from backend.constants import BackendConf
+
+
+async def exporter(data):
```
rgaudin, line 9: This should follow the same convention as everywhere else with a class

rgaudin, line 9: The naming is also incorrect. `exporter` doesn't mean anything. This is a an `OpenzimUploadExporter` and it should be in a dedicated module.

here you should just have the empty interface

backend/src/backend/exporters/kiwix_public_exporter.py
```diff
@@ -0,0 +1,10 @@
+from backend.collections.kiwix_public import KiwixPublicCollection
+from backend.exporters import exporter
+from backend.formatters.kiwixlibraryxml import KiwixLibraryXml
+
+
+class KiwixPublicExporter:
+    async def kiwix_library_xml_exporter():
```
rgaudin, line 7: Also use a generic interface here.

backend/tests/test_exporter.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+@pytest.mark.asyncio
+async def test_kiwix_public_exporter(monkeypatch):
+    async def mock_check_and_upload_file(src_path, upload_uri, private_key, delete):
+        # So as not to upload the server test file
+        ...
+
+    monkeypatch.setattr(
+        openzim_uploader, "check_and_upload_file", mock_check_and_upload_file
```
rgaudin, line 14: You should mock the `upload_file` function that is being called by `check_and_upload_file` so that we can still test the checking parts.

That would mean adding some tests to validate good and bad inputs.

rgaudin (changes requested):
See inline comments.
Also, there's a conflict that you should look into (rebase from master) as nobody else works on this.

I'll add a ticket to add a trigger for this as we'll need to test this live as well

rgaudin (changes requested):
I believe we agreed on testing the `check_and_upload_file` call in the tests like missing file to upload or incorrect `UPLOAD_URI`.

I also notice that there are warnings in the tests output regarding calls that were not awaited. This should be fixed (although it may originate from previous code).

Making sure upload status is reported is important ; see inline comments

rgaudin (changes requested):
See inline comments

backend/tests/test_exporter.py
```diff
@@ -0,0 +1,53 @@
...
+async def test_exporter_interface():
+
+    with pytest.raises(NotImplementedError):
+        for _ in await ExporterInterface().export():
+            ...
+
+
+@pytest.mark.asyncio
+async def test_export(monkeypatch):
+
```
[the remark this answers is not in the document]
rgaudin: sorry

backend/src/backend/exporters/kiwix_public_exporter.py
```diff
@@ -0,0 +1,38 @@
...
+import openzim_uploader
+
+from backend.collections.kiwix_public import KiwixPublicCollection
+from backend.constants import BackendConf, logger
+from backend.exporters import ExporterInterface
+from backend.formatters.kiwixlibraryxml import KiwixLibraryXml
+
+
+class KiwixPublicExporter(ExporterInterface):
+    """Upload XML Kiwix Library to"""
```
rgaudin, line 13: to? 
```py
    """ Upload XML Kiwix Library to `UPLOAD_URI`, using `PRIVATE_KEY`"""
```

Given #92, it might be good to use a more-specific environ and variable name

backend/tests/test_exporter.py
```diff
@@ -0,0 +1,61 @@
...
+    ):
+        return 0
+
+    with monkeypatch.context() as mp:
+        mp.setattr(openzim_uploader, "upload_file", mock_upload_file)
+        mp.setattr(openzim_uploader, "ack_host_fingerprint", mock_ack_host_fingerprint)
+        mp.setattr(BackendConf, "upload_uri", "sftp://localhost:8080/data/test/")
+        mp.setattr(BackendConf, "private_key", ssh_private_key)
+
+        await KiwixPublicExporter.export()
```
rgaudin, line 61: This is not asserted

rgaudin (changes requested):
Thanks ; it's almost there but we need to test that common problems (namely our inputs) are caught, otherwise tests are not as useful:
- incorrect URL should fail
- incorrect RSA key should fail

backend/tests/test_exporter.py
rgaudin: I don't see it being resolved…

backend/tests/test_exporter.py
```diff
@@ -0,0 +1,61 @@
...
+    def mock_named_temporary_file(delete=False):
+        class MockNamedTemporaryFile(object):
+            def __init__(self, *args, **kwargs):
+                self.name = "/fakefile"
+
+        return MockNamedTemporaryFile()
+
+    with monkeypatch.context() as mp:
+        mp.setattr(tempfile, "NamedTemporaryFile", mock_named_temporary_file)
+        assert await KiwixPublicExporter.export() == 0
```
rgaudin, line 30: Misleading `== 0` while this should return `True` or `False`. This means it failed

fadiga: OK

backend/tests/test_exporter.py
```diff
@@ -0,0 +1,75 @@
...
+async def test_incorrect_rsa_key(monkeypatch, ssh_private_key):
+    with monkeypatch.context() as mp:
+        mp.setattr(BackendConf, "private_key", ssh_private_key)
+    assert not await KiwixPublicExporter.export()
+
+
+@pytest.mark.asyncio
+async def test_incorrect_url(monkeypatch):
+    with monkeypatch.context() as mp:
+        mp.setattr(BackendConf, "upload_uri", "sftp://localhost:8080/data/test/")
```
rgaudin, line 43: Your test is named `test_incorrect_url` yet the URL you are testing with in the test is identical to the one you are testing with in the next test which is a successful *upload*. Something is not right.

rgaudin (changes requested):
See inline comment ; either the test or its name is incorrect

rgaudin (approved):

PR #93, opened 2022-10-03
Added export button

### LINKED ISSUE #91: Add export trigger
We need a button to trigger the kiwix xml exporter to test both the feature and the dataset

fadiga:
close #91


rgaudin (approved):
Looks good. This is temp feature that will be removed in a close future. That's why it's separate and not tested.

PR #98, opened 2023-10-04
Fix language query to fetch only based on codename

benoit74:
# Rationale
Probably due to an upstream change (in Babel ?), languages `name` and `native` properties are not up-to-date anymore in the DB. Code was not properly handling this situation, probably assuming these properties will never change.

# Changes
- Language objects are fetched only based on their `code`
- Other language objects properties (`name` and `native`) are updated if they have changed


rgaudin (approved):
Makes sense ; thank you.
What's with the formatting changes?

benoit74:
Regarding formatting, looks like black has changed some convention ^^

rgaudin:
> Regarding formatting, looks like black has changed some convention ^^

But the workflow checks for black on a pinned version…

benoit74:
My fault, looks like I did not used the local version but my system one ...

rgaudin:
Passed the QA check anyway 👍

PR #138, opened 2026-02-03
secure POST endpoints with authentication

### LINKED ISSUE #134: Secure "write" API endpoints with authentication based on Ory.sh
We need to implement authentication on the CMS to secure "write" endpoints:
- endpoint called by the Zimfarm to notify the CMS
- endpoints called by "editors" to modify CMS configuration (create title, alter title configuration, ...)

For this we need a user table in the DB with associated username + Ory `sub` + a role column.

For now, we will have only three roles:
- `zimfarm` for the zimfarm, authorized only to send notifications
- `editor` for users, authorized to do "anything"
- `viewer` for new users, same as "unauthenticated"

We need UI to let users sign-in on Ory, or sign-out.

Just like on Zimfarm, any new user (authenticated by Ory) is automatically added to CMS DB with a `viewer` role.

We do not need UI views to handle users for now, this will be tracked in a distinct issue. Admin will take care of setting the role manually directly in DB if needed.

Each "write" endpoint will check if current user has sufficient permissions to perform the API operation.

elfkuzco:
## Rationale
This PR secures POST endpoints with functions that check user has access to resource. It also adds authentication via Kiwix login (handled by Ory) on the frontend

## Changes
- add roles `VIEWER`, `EDITOR` and `ZIMFARM` with `EDITOR` being the user that is allowed to do anything on the project
- create generic permission `ResourcePermission` for CRUD operations. Editor has it set for all operations. `ZIMFARM` has only the zimfarm resource  create and read permissions. This allows to cleanly use a dependency function on the route itself rather than doing checks within the route.
- set up authentication with Kiwix login on frontend
- use dependency function to restrict acess to `POST` endpoints (only write endpoints for now)

This closes #134


backend/src/cms_backend/api/routes/dependencies.py
```diff
@@ -0,0 +1,123 @@
...
+]
+
+
+def get_jwt_claims_or_none(
+    authorization: AuthorizationCredentials,
+) -> JWTClaims | None:
+    """
+    Get the JWT claims or None if the user is not authenticated.
+
+    Tries to decode as Zimfarm token first, then as Kiwix token.
```
benoit74: There is no such thing as "Zimfarm token" for now

benoit74 (changes requested):
Thank you, well done! It works mostly like a charm, I like the fact the experience you've gathered on Zimfarm is so smooth to deploy here.

Two "minor" things to fix and we are good to merge:
- the "Sign-in" button should switch to the display of username + on click a drop-down with "copy token" and "sign-out" (just like on Zimfarm, but without the "change passord" since it makes no sense)
- the "Create Title" button should be hidden when role is not "editor"

PR #148, opened 2026-02-06
create title with collection name and path

### LINKED ISSUE #139: Enhance the "Create Title" popup
The "Create Title" popup in the UI needs to be enhanced to:
- propose only known maturities (`dev` and `robust` for now, see https://github.com/openzim/cms/issues/129) in checkboxes
- allow to setup in which collection we want to publish the ZIM, and for each collections in which subfolder
- for that to work, we need to 
  - store in DB `collection` table the list of allowed subfolders (empty subfolder `""`, meaning use the root folder, must be allowed to be used) ; we can probably use an array column type
  - provide an API endpoint to retrieve the list of all collections and their respective subfolders

Image

Nota: 
- configuration of the collections does not has to be covered in this issue
- do not forget to fix `setup_collection.py` to also populate the new DB column

elfkuzco:
## Rationale
This PR enhances the API/UI to create titles with a collection name and path
Screenshot_20260206_104332
Screenshot_20260206_104322

## Changes
- add endpoint to retrieve a list of collections
- enhance model schema to create title so that values are `NotEmptyString`s
- add methods to retrieve the collection by name
- remove occurences of using cookies to store user-preferences
- add store for collections

This closes #139

### DIFF
--- a/backend/src/cms_backend/api/routes/collection.py
+++ b/backend/src/cms_backend/api/routes/collection.py
@@ -3,20 +3,50 @@ from typing import Annotated
 from uuid import UUID
 from xml.etree import ElementTree as ET
 
-from fastapi import APIRouter, Depends, Path
+from fastapi import APIRouter, Depends, Path, Query
 from fastapi.responses import Response
 from sqlalchemy.orm import Session as OrmSession
 
+from cms_backend.api.routes.fields import LimitFieldMax200, SkipField
+from cms_backend.api.routes.models import ListResponse, calculate_pagination_metadata
 from cms_backend.db import gen_dbsession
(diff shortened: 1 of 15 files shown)


benoit74 (changes requested):
We are getting close, one last requirement which was unclear obviously: 1 title can be associated with at most 1 collection. I.e. we need to detect duplicates on collection only, not collection + path. We cannot have a title association with `collection1::path1` and with `collection1:path2`. We can only have `collection1::path1` and `collection2::path2`. To be done both in backend and frontend obviously.

benoit74 (commented):
One minor wording fix, you can merge directly once applied, LGTM

PR #149, opened 2026-02-09
schedule old production books to be deleted

### LINKED ISSUE #141: Automatically delete old production books
Whenever a new book arrives in a title with `location_kind` set to `prod`, the `mill` processing engine needs to decide if some older books needs to be deleted.

The retention rules are described in https://wiki.openzim.org/wiki/ZIM_Updates:

```
- keep the last version of two ZIM files from the two last distinct months (e.g. if we have `2024-04`, `2024-04a`, `2024-06`, `2024-06a`, `2024-06b`, then we keep `2024-04a` and `2024-06b` ; hopefully this is a scenario which will never happen, but this is the algorithm)
- AND keep every version which is 30 days old or less
```

It has to be understood as:
- group `prod` books by title and flavour
- for each group, apply the retention rules

So when a title arrives in the `prod` of a given title, we apply this logic.

When a book has to be deleted, we do not want to remove it immediately from the filesystem because we might have ongoing downloads and the library is still using it. Logic is hence to first set its `location_kind` to `to_delete` and the `deletion_date` is set. `deletion_date` is a new `book` table column to add. It is computed based on current datetime + a given elapse (configurable via environment variable, default to `1d`)

On the shuttle side, processing has to be adapted to delete these books from filesystems once current date is after `deletion_date`. Once file is really deleted from filesystem, the book `location_kind` is set to `deleted` (and `deletion_date` is kept as-is).

Book details view has to be adapted to show the new `deletion_date` field.

elfkuzco:
## Rationale
This PR enhances the `mill` to apply [retention rules](https://wiki.openzim.org/wiki/ZIM_Updates) to `prod` titles. It groups the `prod` titles by flavour before applying the rules to determine which files to mark as deleted. Consequently, a new background task is added to the `shuttle` which actually deletes the files that have been marked for deletion

## Chanages
- add new column `deletion_date` to set the date the file should be deleted
- apply retention rules to files
- add background task to delete files that have been marked for deletion

This closes #141


benoit74 (changes requested):
Looks pretty good, I just think there is a bug in book sorting (unless I'm mistaken, but this is not tested so better to at least fix the test to cover it, probably fix the code). Situation to consider is that we could have multiple letters in the period, like `2021_01aa`. And `2021_01a <2021_01b < 2021_01c < ... < 2021_01aa < 2021_01bg` which is probably not what happens with the natural alphabetical sorting the code is using. Having two letters suffix is going to be extremely rare, but we need to take that into account. Sort expression which is already used in `compute_target_filename` is `suffixes.sort(key=lambda s: (len(s), s))`

PR #151, opened 2026-02-10
setup local login on cms

### LINKED ISSUE #146: Add ability to login with local username + password
https://github.com/openzim/cms/pull/138 has introduced user login with central Ory.sh Kiwix instance.

While we do not intend to use it for production, being able to login with username and password would be useful for:
- development (use Ory.sh login is a bit cumbersome since we need to expose app on the web, allow app URL as redirect URL in ory.sh which is an issue for external contributors, ...)
- production fallback should Ory.sh be unavailable

We need to add a `password` column in the DB `user` table. We still do not want UI screens to manage users or user screen to change his own password, this will be a distinct issue.

We need UI configuration flag to decide what we show in the sign-in screen: local username+password and/or Kiwix login.

@elfkuzco how big / complex would it be to implement this? Is it worth it to do it now or should we wait a bit since we are anyway only two developers and production fallback is kind of a "nightmare" situation we can live without?

elfkuzco:
## Rationale
This PR enhances the CMS by adding the functionality required to register/sign in users with a local authentication mechanism (password + username combination)

## Changes
- add `password_hash` to user model, make `id` column primary key instead of `idp_sub`
- add table for creating/deleting refresh tokens
- set up local authentication provider on frontend
- use local tokens for testing

This closes #146


elfkuzco:
Going to need #149  review before merging as it's directly based on it because of the migrations they both introduce

PR #163, opened 2026-02-16
fix maint-scripts/populate_cms_database.py script issues

### LINKED ISSUE #162: Maint-script: few issues detected
- CMS `book` table should use `bigint` for size, `int` is too small ; it is probably safe to also change `article_count` and `media_count` to `bigint`, pretty sure we do not yet have such big ZIMs, but it might appear at some point and CMS should not be a blocker.

- logic in `find_collections_for_zim` is wrong, it returns one entry per collection title (multiple times the same collection) ; we can fix it with something like this:

```
    for collection in collections:
        if (any([1 if collection_title.path == zim_dir else 0 for collection_title in collection.titles])):
            results.append(collection)
```

- we need to commit to DB after every ZIM processed, so that we do not reprocess same stuff over and over again

elfkuzco:
## Changes
- change type of `media_count`, `article_count` and `size` from `int` to `BigInteger` for `Book` table
- refactor logic to find collections for a zim
- set `ALEMBIC_UPGRADE_HEAD_ON_START` for api in dev compose file

This closes #162


benoit74 (approved):
LGTM

PR #165, opened 2026-02-16
add tests for mill and shuttle components on ci

elfkuzco:
## Changes
- use argparse to parse arguments from users for mill and shuttle CLI
- use commands for invoking scripts in Dockerfile instead of relying on `if __name__` check and invoking interpreter as `pyproject.toml` already specifies scripts
- add tests for shuttle and milll components. The `--help` flag exits but has the side effect of importing everything the components need to run
- enhance API tests to fail if component doesn't respond on time


benoit74 (approved):
LGTM

PR #170, opened 2026-02-17
fix catalog generation and populate zims using id from info

### LINKED ISSUE #166: Small fixes following initial DB loading
Initial DB loading with all production ZIMs (not dev/staging ones) gave very promising results.

Catalog (full XML) achieves to load in 1 to 2 seconds, directly from DB.

I've identified some small bugs:
- in `populate_cms_database` maintenance script, we create books with a random UUID while their ID must be the ZIM UUID (as retrieved from ZIM metadata)
- the catalog response is one a single line (no carriage return) while we used to have one book per line, with carriage returns, see https://download.kiwix.org/library/library_zim.xml ; this was pretty convenient ; not sure it is easy to fix but having everything on one line is definitely not convenient at all
- the catalog response misses following keys:
  - `faviconMimeType`: must be `image/png`
  - `favicon`: base64-encoded ZIM illustration
  - `url`: should be something like `https://download.kiwix.org/zim/stack_exchange/3dprinting.stackexchange.com_en_all_2026-02.zim.meta4` (mind the `.meta4` at the end)
- the catalog response has following issues:
  - `size` must be in KiB (divide current size by 1024)
  - `tags` misses some values "sometimes"

Regarding missing tags:
- if we focus on this task : https://farm.openzim.org/pipeline/f34fe28d-d97f-4a1a-b404-31c96809c681
- old library XML advertise this with `stack_exchange;_category:stack_exchange;_details:no;_videos:no;_ftindex:no;_pictures:yes` tags while on Zimfarm and on CMS side we see only `stack_exchange;_category:stack_exchange;_details:no;_videos:no`
- this is all linked to the https://github.com/openzim/python-scraperlib/blob/99047776f5998ef4244a78afb5499e2c23890c70/src/zimscraperlib/zim/_libkiwix.py#L92 conversion ; we need to do this as well in the CMS, only when generating the catalog for now

elfkuzco:
## Changes
- add helper function to retrieve missing keys from any dictionary and use to validate missing info keys and missing metdata keys from zim info
- convert tags using scraperlib logic
- divide sizes by 1024
- indent catalog xml output

This fixes #166

### DIFF
--- a/backend/src/cms_backend/api/routes/collection.py
+++ b/backend/src/cms_backend/api/routes/collection.py
@@ -1,3 +1,4 @@
+import math
 from http import HTTPStatus
 from typing import Annotated
 from uuid import UUID
@@ -11,15 +12,17 @@ from cms_backend.api.routes.fields import LimitFieldMax200, SkipField
 from cms_backend.api.routes.models import ListResponse, calculate_pagination_metadata
 from cms_backend.db import gen_dbsession
 from cms_backend.db.collection import (
+    CollectionBook,
     get_collection,
     get_collection_by_name_or_none,
     get_latest_books_for_collection,
 )
 from cms_backend.db.collection import get_collections as db_get_collections
 from cms_backend.db.exceptions import RecordDoesNotExistError
-from cms_backend.db.models import Book
 from cms_backend.schemas import BaseModel
 from cms_backend.schemas.orms import CollectionLightSchema
+from cms_backend.utils.filename import construct_download_url
+from cms_backend.utils.zim import convert_tags
 
 router = APIRouter(prefix="/collections", tags=["collections"])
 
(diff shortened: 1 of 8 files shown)


backend/src/cms_backend/api/routes/collection.py
```diff
@@ -62,7 +63,7 @@ def _build_library_xml(books: list[Book]) -> str:
 
         # Required attributes
         book_elem.set("id", str(book.id))
-        book_elem.set("size", str(book.size))
+        book_elem.set("size", str(book.size / 1024 if book.size > 0 else book.size))
```
benoit74: We need to round to nearest integer.

backend/tests/api/routes/test_library.py
```diff
@@ -192,7 +192,7 @@ def test_get_collection_catalog_xml_single_book(
     assert book_elem.get("publisher") == "Test Publisher"
     assert book_elem.get("name") == "test_title"
     assert book_elem.get("date") == "2025-01-01"
-    assert book_elem.get("size") == str(book.size)
+    assert book_elem.get("size") == str(book.size / 1024)
     assert book_elem.get("mediaCount") == str(book.media_count)
     assert book_elem.get("articleCount") == str(book.article_count)
     assert book_elem.get("tags") == "_category:test;_pictures:yes"
```
benoit74: Why don't we have `_ftindex`, `_details` and `_videos` tags here, based on the new "conversion" logic?

PR #171, opened 2026-02-18
decode oauth2 client generated tokens

elfkuzco:
## Rationale
This PR adds a new token decoder for decoding oauth client generated tokens using client_id and client_secret. Such tokens will come from the zimfarm in order to create notifications on the CSM

## Changes
- add client_id env variable to ensure token client_id matches configured one
- add decoder to chain of decoders

### DIFF
--- a/backend/src/cms_backend/api/context.py
+++ b/backend/src/cms_backend/api/context.py
@@ -24,13 +24,18 @@ class Context:
     oauth_session_login_require_2fa = parse_bool(
         os.getenv("OAUTH_SESSION_LOGIN_REQUIRE_2FA", default="true")
     )
+    oauth_client_id = os.getenv(
+        "OAUTH_CLIENT_ID", default="310c5189-ce06-463c-9c55-46e822b5d642"
+    )
     create_new_oauth_account = parse_bool(
         os.getenv("CREATE_NEW_OAUTH_ACCOUNT", default="true")
     )
-    # List of authentication modes. Allowed values are "local", "oauth-session"
+    # List of authentication modes. Allowed values are
+    # - local
+    # - oauth
     auth_modes: list[str] = os.getenv(
         "AUTH_MODES",
-        default="oauth-session",
+        default="oauth",
     ).split(",")
 
     # Local Authentication JWT settings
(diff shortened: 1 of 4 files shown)


benoit74 (commented):
I don't get why we would need a new decoder.

JWT payload (from the test I just ran) is like this:

```
{
  "aud": [],
  "client_id": "310c5189-ce06-463c-9c55-46e822b5d642",
  "exp": 1771500664,
  "ext": {},
  "iat": 1771497064,
  "iss": "https://login-staging.kiwix.org",
  "jti": "c01a713c-834f-431a-9327-e83fe2ea4c7e",
  "nbf": 1771497064,
  "scp": [],
  "sub": "310c5189-ce06-463c-9c55-46e822b5d642"
}
```

`client_id ` is hence already present in the `sub` ; forcing a specific client ID prevents us to have other service authenticating to the CMS (should we need to, and we might need it).

Only problem when comparing with an `oauth-session` token is that it misses the `aud` value and has no `aal` and `name` property.

```
{
  "aal": "aal2",
  "aud": [
    "bada4130-5143-4524-a0bb-0d69671beee2"
  ],
  "exp": 1771498614,
  "iat": 1771498014,
  "iss": "https://login.kiwix.org",
  "jti": "e805398a-6938-4920-a923-413c7c8c3c7c",
  "name": "Benoît",
  "nbf": 1771498014,
  "sid": "9907a04f-6857-4623-a062-27d051db73a4",
  "sub": "beadca2c-f960-4c80-8b68-3245353584b6"
}
```

I've just modified Ory configuration and it is now possible to specify `grant_type=client_credentials&audience=309693e7-ad5e-4379-bf93-ba89314230fd` (for staging) to get proper `aud` value in the JWT.

Wouldn't it be simpler to modify `oauth-session` decode to make it more generic and:
- request `aal` = `aal2` only if `client_id` is not present
- ensure `client_id` is same than `sub`

Then it would automatically map to the user with matching `sub` as "usual" to get its rights.

Probably means we need to rename `oauth-session` to `oauth`, but I'm fine with this change.

Or do I miss something?

PR #201, opened 2026-03-05
process events from Redis streams

elfkuzco:
## Rationale
This PR leverages [Redis Streams](https://redis.io/docs/latest/develop/data-types/streams) and [consumer groups](https://redis.io/docs/latest/develop/data-types/streams/#consumer-groups) to publish events when a title is created or modified. These events are picked up by a processor which acknowledges the events and attaches books that match the title to the new title.

## Changes
- publish stream events when a title is created or modified
- add processor to read, process and acknowledge events. Redis streams and consumer groups allow us to recover from failures/crashes and re-process messages which we haven't acknowledged.
- add env variable for setting the name of the consumer. This should allow one to spin up different instances with different consumer names. This is also the reason why the streams processor is added as a standalone script instead of another background task.

### DIFF
--- a/backend/src/cms_backend/redis/consumer.py
+++ b/backend/src/cms_backend/redis/consumer.py
@@ -63,7 +63,7 @@ class RedisConsumer:
         try:
             # Check for pending messages (from this consumer)
             if self._check_backlog:
-                stream_id = "0"  # '0' means my pending messages
+                stream_id = "0-0"  # '0-0' means my pending messages
             else:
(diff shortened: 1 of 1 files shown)


elfkuzco:
Okay. My plan initially was to create an `Event` table and then when a title is created/modified, the event would be created. The background task would notice it (perhaps when next it runs as a background task), act on it and then, delete processed events from the DB so it doesn't get filled up with useless entries. WDYT of this approach?

PR #204, opened 2026-03-10
move and delete books manually

### LINKED ISSUE #143: Allow to move and delete books manually
API needs to have endpoints + the book details UI needs to have buttons to:
- allow to mark any book for deletion ; same delay applies before really deleting the book as in https://github.com/openzim/cms/issues/141 (even if not for the exact same reasons, here it is more to allow rollback if needed)
- allow to move books in `staging` to `prod`: the `location_kind` is updated to `prod`, `file_operation_needed` flag is set and target locations are computed ; move will happen mostly immediately
- allow to move books in `prod` to `staging`: the `location_kind` is updated to `staging`, `file_operation_needed` flag is set and target locations are computed ; move will happen mostly immediately

For both operations, we need a UI popup asking user to confirm the operation. When moving book from `prod` to `staging` a special warning should say `This move will happen immediately. Any user downloading the ZIM, using it in the imager service or accessing it from the library will have a broken experience. Use only in extremely controlled events.`

elfkuzco:
## Rationale
This PR adds functionality to move, recover and delete books from the API/UI. 
Screenshot_20260310_045108
Screenshot_20260310_045044
Screenshot_20260310_045027

## Changes
- add endpoints to delete, move and recover book
- move quarantine and staging related warehouse env vars from mill to common context as they are now used by the API layer too
- enhance `get_book_or_none` function to allow for passing params for filtering instead of having `if...else` checks for attributes
- add db functions to move, recover and delete book
- previously, the shuttle `delete_files` logic set the book `has_error` if deletion failed. This doesn't seem correct as the book does not necessarily have an error. Instead, in addition to logging the exception message, set the `needs_file_operation` to False
- remove the `has_error` flag from filters to get_next_book_to_delete as we want to now delete books that have errors too
- add UI buttons to move, recover and delete book
- only include `quarantine`, `staging` and/or `prod` books in library.xml generation (filter added in DB query `get_latest_books_for_collection`)

This closes #143


benoit74 (changes requested):
Very good, thank you. I've identified few small flaws and I will open few more issues for stuff not yet identified as necessary for a smooth user experience.

Fix book status when book is pending deletion (both in book details and title details views), we should not have "Unknown" shown here:

Image

Also when I mark a book for immediate "force deletion", the "recover" button is still shown (this might be acceptable) but the API call fails (this is expected) and the message says "Book ... is not eligible for deletion" (this is incorrect)

Image

Display location_kind next to the green checkmark (in addition to the tooltip, both in book details and title details views):

Image

When book is deleted, status is back to "Published" (in the UI) while it should be "Deleted" (both in book details and title details view)

Image

PR #224, opened 2026-04-15
compress responses using gzip middleware

### LINKED ISSUE #223: gzip the XML results of https://api.cms.openzim.org/v1/collections/{xxx}/catalog.xml
E.g. https://api.cms.openzim.org/v1/collections/Kiwix/catalog.xml

The file is "pretty big", so it should be gziped to save bandwidth (if reasonably feasible).

elfkuzco:
## Changes
- add gzip middleware to app to compress responses

This closes #223

### DIFF
--- a/backend/src/cms_backend/api/main.py
+++ b/backend/src/cms_backend/api/main.py
@@ -6,6 +6,7 @@ from http import HTTPStatus
 from fastapi import APIRouter, FastAPI, HTTPException
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.middleware.gzip import GZipMiddleware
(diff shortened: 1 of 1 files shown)


kelson42:
Why not supporting zstd, which is significantly better and pretty lightweigth?

elfkuzco:
Okay. Didn't know about it. Doesn't come with FastAPI by default. Will look for a tested library or implement one

kelson42:
If it does not come in standard, then I would recommend to leave it for the moment.

elfkuzco:
> If it does not come in standard, then I would recommend to leave it for the moment.

Here's one I tried before seeing this comment: https://github.com/tuffnatty/zstd-asgi

benoit74 (approved):
LGTM

Issue #225, opened 2026-04-16
Add support for having a kiwix-serve on staging area


benoit74:
We wanna deploy a kiwix-serve on top of CMS staging area (what is currently seen at https://dev.library.kiwix.org).

This means that we need to:
- add support getting staging `library.xml` file with proper download locations (beware the specificity on item url linked to the fact we do not have a mirrorbrain running on dev)
- add support for getting ZIM URLs (download and view) even when ZIM is in staging
- obviously we need a configuration for download and view base URLs for staging

PR #226, opened 2026-04-16
Library for collections must consider only books in production

benoit74:
This PR ensures we do not have a subtle bug in library.xml generation. 

This is rather defensive coding because since staging and quarantine are using specific paths, we should never have a record in DB stored in the collection path but not in "prod" status. But still, the condition was not exactly correct and I was worried of any consequence when working on https://github.com/openzim/cms/issues/225


elfkuzco (approved):
LGTM

PR #227, opened 2026-04-17
Expose staging books

### LINKED ISSUE #225: Add support for having a kiwix-serve on staging area

benoit74:
Fix #225 

Side-effect: conventions for base URLs (download and view) have been clarified (and slightly changed) to be more flexible (not necessary needed today, but made quite visible by staging situation) ; I will fix DB as required


elfkuzco (approved):
LGTM

PR #238, opened 2026-04-24
include/exclude staging and quarantine books if needs_attention

### LINKED ISSUE #205: Display quarantine, staging, needs_file_operation and has_error books in inbox
In the Inbox tab, we want to display all books which require user attention. 

This means quarantine (as of today) and staging (to be added) books.

Also display there books which needs a file operation (even if they do not require user action, but having them centralized in a single view is convenient) and books which have errors.

Books should be sorted for better readability:
- first by status: `has_error` then `quarantine` then `staging` then `needs_file_operation`
- then by `created_at` (descending, i.e. most recent books first)

elfkuzco:
## Changes
- include books in staging/quarantine if `needs_condition` is True
- exclude books in `staging/quarantine` if needs_condition is False

This closes #205

### DIFF
--- a/backend/src/cms_backend/db/books.py
+++ b/backend/src/cms_backend/db/books.py
@@ -66,6 +66,7 @@ def get_books(
     if needs_attention is True:
         stmt = stmt.where(
             or_(
+                Book.location_kind.in_(["quarantine", "staging"]),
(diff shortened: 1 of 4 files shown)


benoit74 (approved):
LGTM

PR #239, opened 2026-04-24
sort books by multiple columns

### LINKED ISSUE #205: Display quarantine, staging, needs_file_operation and has_error books in inbox

elfkuzco:
## Rationale

This PR implements the sorting order described in https://github.com/openzim/cms/issues/205#issue-4050909565.

Books are sorted by:
- `has_error` (truthy values come first)
- `location_kind`: alpahabetical sort, thus, quarantine comes before staging
- `needs_file_operation`: (truthy values come first)
- `created_at`: in descending order

This closes #205

### DIFF
--- a/backend/src/cms_backend/db/books.py
+++ b/backend/src/cms_backend/db/books.py
@@ -40,7 +40,13 @@ def get_books(
         Book.name,
         Book.date,
         Book.flavour,
-    ).order_by(Book.id)
+    ).order_by(
+        Book.has_error.desc(),
+        Book.location_kind,
+        Book.needs_file_operation.desc(),
+        Book.created_at.desc(),
+        Book.id,
+    )
 
     if book_id is not None:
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM ; not sure it will work "forever" once we have more location_kin included in the screen, but sufficient for the time being, so let's work with this for now.

elfkuzco:
Do you think these location kinds should be an enum along with other values that are currently magic strings?

PR #248, opened 2026-04-27
compute etags for collection catalog xml response

### LINKED ISSUE #237: Support ETag for catalog XML endpoints
For library.kiwix.org and dev.library.kiwix.org, we are polling the `/collections/Kiwix/catalog.xml` and `/staging/catalog.xml` endpoints (there is no push mechanism) frequently to retrieve the catalog.

If those responses had an ETag, we'd save a lot of requests and more importantly, a lot of reloads (kiwix-serve, varnish).

I don't know the code base but I guess a checksum of the books's ID set could be computed easily.

Catalog-retriever already supports both mode (with or without etag).

elfkuzco:
## Rationale

This PR computes the hash of the XML response of the collections catalog and adds it in the response headers of the request both for the `GET` and `HEAD` requests

## Changes
- add endpoint to compute hash of XML response (accessible via `HEAD` requests)
- add `Etag` header in `GET` requests. Header value is the hash of the XML response

This closes #237

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
     "regex == 2025.10.23",
     "humanfriendly == 10.0",
     "Werkzeug == 3.1.5",
+    "xxhash == 3.7.0",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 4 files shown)


benoit74 (changes requested):
LGTM, except that we need the same for special `/staging/catalog.xml` endpoint (which is implemented in dedicated router since it is not same path).

benoit74 (approved):
LGTM

PR #256, opened 2026-04-28
sort books in title details by date and created_at

### LINKED ISSUE #235: Sort books in title details by "Date" then "Created"
The list of books in title details must be sorted by "Date" (coming from the ZIM file) then "Created". Both descending.

In general it produces the same result, but not always because only "Date" is a true ZIM property, "Created" is the date where book has been created in the CMS which can be very different from "Date" (especially right now where books have been all created at once).

elfkuzco:
# Rationale
This PR sorts books in title details by the combination of the `Date` and `Created At` properties (both in descending order)

## Changes
- select books in load while fetching title
- sort title books by `date` (defaulting to UNIX epoch if not exists) and `created_at`

This closes #235

### DIFF
--- a/backend/src/cms_backend/db/title.py
+++ b/backend/src/cms_backend/db/title.py
@@ -1,3 +1,4 @@
+import datetime
 from pathlib import Path
 from uuid import UUID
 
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #257, opened 2026-04-28
raise conflict errors on title update with existing name

### LINKED ISSUE #250: CMS should not return 500 but 400 when updating with already used title
When I edit a title and choose an already-used title name, API returns a 500 error with traceback while it should be a `400 Bad Request` with details about the conflicting key.

UI should also correctly display the error.

elfkuzco:
## Rationale
This PR catches integrity errors when title name is updated with another title's name
Screenshot_20260428_105037

## Changes
- raise `RecordAlreadyExistsError` if integrity error is raised during title name update
- pass Context debug mode to app during creation to avoid throwing traceback errors to clients

This closes #250

### DIFF
--- a/backend/src/cms_backend/api/main.py
+++ b/backend/src/cms_backend/api/main.py
@@ -80,7 +80,7 @@ def create_app(*, debug: bool = True):
     return app
 
 
-app = create_app()
+app = create_app(debug=Context.debug)
 
 
 @app.exception_handler(RequestValidationError)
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM ; I'm not sure it is the only cause of integrity error we will have when updating a title, but this assumption seems to hold true ATM so let's start with this simplification for now.

PR #258, opened 2026-04-28
apply retention rules using Date Metadata

### LINKED ISSUE #245: CMS does not delete old books
See e.g. https://cms.openzim.org/title/ted_mul_buddhism

Image

Book from `2025-10-27` should now have been deleted. Looks like `apply_retention_rules` either did not triggered or is buggy.

elfkuzco:
## Changes
- apply retention rules using `Date` metadata property instead of `created_at`

This fixes #245

### DIFF
--- a/backend/src/cms_backend/mill/context.py
+++ b/backend/src/cms_backend/mill/context.py
@@ -31,3 +31,9 @@ class Context:
             os.getenv("PROCESS_TILTE_MODIFICATIONS_INTERVAL", default="1m")
         )
     )
+
+    process_retention_rules_interval: timedelta = timedelta(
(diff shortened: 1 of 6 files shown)


benoit74 (changes requested):
LGTM, except that we miss a solution to recover from current DB situation where retention rules did not apply. How should we do that? A one-shot maintenance script? Or should we move to another approach where we delete old books in another background job in the mill?

elfkuzco:
Maintenance script works for me.

elfkuzco:
Added a background job instead in the mill with configurable interval.

benoit74 (approved):
LGTM

PR #259, opened 2026-04-28
apply retention rules during manual move of book from staging to prod

### LINKED ISSUE #246: CMS should delete all books when moving book from staging to prod manually
Looking into #245, I realized that we do not `apply_retention_rules` when moving a book from staging to prod, while we should. all books in `dev` maturity (and maybe more with next CMS modifications) will go through staging for manual inspection before moving to prod, and these will need to trigger older book deletions as well.

elfkuzco:
## Rationale
This PR applies the retention rules to the books belonging to a title when a book is moved from staging to prod manually

## Changes
- add new module `book_location.py` for keeping book location related logic and avoid circular imports
- move retention rules to `db/title.py` module
- apply retention rules if book does not go to staging

This closes #246

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -1,5 +1,3 @@
-from dataclasses import dataclass
-from pathlib import Path
 from typing import Any, Literal
 from uuid import UUID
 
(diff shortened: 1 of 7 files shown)


benoit74 (approved):
LGTM

PR #260, opened 2026-04-28
Permissive healthcheck

### LINKED ISSUE #231: Healthcheck should be more permissive
The "Zimfarm Notification", "File Needs Move" and "File Needs Deletion" are too strict. They raise immediately while everything is async.

E.g. it is normal and expected to have zimfarm notifications pending to process for few minutes, until the mill triggers and process these notifications into books. Same for the two other checks.

We need to make these checks more permissive, e.g. raise an alarm regarding zimfarm notifications only if notifications is pending since too long.

elfkuzco:
## Rationale
This PR makes the healthcheck to filter books and notifications on pending state for too long either via their `updated_at` or `received_at` properties respectively. 

## Changes
- add `updated_at` column to book model with `onupdate` set to `getnow`. This way, any actions performed on book will automatically update the `updated_at` column
- filter books and notifications requiring action using their timestamp as CMS actions might take time before kicking off

This fixes #231


benoit74 (approved):
LGTM

PR #261, opened 2026-04-29
extend default deletion delay

elfkuzco:
## Changes
- change default deletion delay for healthcheck to be more than CMS deletion delay. This is because it should only be an issue if the book is still marked as `to_delete` after CMS deletion delay has expired


benoit74 (approved):
LGTM

PR #262, opened 2026-04-29
query db for latest books during zim URL computation

### LINKED ISSUE #253: Computation of ZIM to view has a bug when title changed
See https://cms.openzim.org/title/madrasa_ar_electeng

Image

ZIM to expose in the catalog and which should have a view URL should be the one from `2026-02-27`, not the one from `2024-03-04`.

I'm quite sure this is somehow linked to the fact that ZIM name has changed.

The good news is that the catalog.xml and the `/books/zims` API are consistent. They just choose the wrong book.

We should probably have a look as well at the logic which decides which books should be deleted to confirm it is not impacted.

### LINKED ISSUE #254: Querying for a single ZIM URLs gives wrong results
Querying for two ZIMs (latest one and the one before on a single title) gives expected results:

https://api.cms.openzim.org/v1/books/zims?zim_ids=8ec5b0b8-a8ce-acd9-e414-5a97d8187617&zim_ids=0ff05730-3ead-f792-b942-f0884feb70c2

```json
{
  "urls": {
    "8ec5b0b8-a8ce-acd9-e414-5a97d8187617": [
      {
        "kind": "download",
        "url": "https://download.kiwix.org/zim/ted/ted_mul_diversity_2026-01.zim",
        "collection": "Kiwix"
      }
    ],
    "0ff05730-3ead-f792-b942-f0884feb70c2": [
      {
        "kind": "download",
        "url": "https://download.kiwix.org/zim/ted/ted_mul_diversity_2026-04.zim",
        "collection": "Kiwix"
      },
      {
        "kind": "view",
        "url": "https://browse.library.kiwix.org/viewer#ted_mul_diversity_2026-04",
        "collection": "Kiwix"
      }
    ]
  }
}
```

Querying for a single ZIM does not work as expected: it should give "consistent" (and accurate) results without a view link:

Currently:

https://api.cms.openzim.org/v1/books/zims?zim_ids=8ec5b0b8-a8ce-acd9-e414-5a97d8187617

```json
{
  "urls": {
    "8ec5b0b8-a8ce-acd9-e414-5a97d8187617": [
      {
        "kind": "download",
        "url": "https://download.kiwix.org/zim/ted/ted_mul_diversity_2026-01.zim",
        "collection": "Kiwix"
      },
      {
        "kind": "view",
        "url": "https://browse.library.kiwix.org/viewer#ted_mul_diversity_2026-01",
        "collection": "Kiwix"
      }
    ]
  }
}
```

Expected:

```json
{
  "urls": {
    "8ec5b0b8-a8ce-acd9-e414-5a97d8187617": [
      {
        "kind": "download",
        "url": "https://download.kiwix.org/zim/ted/ted_mul_diversity_2026-01.zim",
        "collection": "Kiwix"
      }
    ]
  }
}
```

elfkuzco:
## Rationale
This PR refactors the way zim URLs are computed by using a subquery to get the books which are the latest per title/flavor combination. It then joins this subquery to the original query and building up a column `is_latest` if the current book date matches the one from the latest subquery. This way, we can/cannot have a view URL depending on if the latest book is included in the book IDs to compute their URL.

## Changes
- build up the list of titles for all the book IDs given
- use subquery to get the latest books for the titles
- join original query to the subquery since all the books will always have one title_id/flavor. This is necessary to build up the `is_latest` property
- sort books by zim metadata `Date` first before `created_at`. This will fix the issue in #253 

This fixes #253 
This fixes #254

### DIFF
--- a/backend/src/cms_backend/api/token.py
+++ b/backend/src/cms_backend/api/token.py
@@ -155,10 +155,9 @@ class OAuthTokenDecoder(TokenDecoder):
         except Exception:
             return False
 
-        if payload.get(
-            "iss"
-        ) != Context.oauth_issuer or Context.oauth_session_audience_id not in payload.get(
-            "aud", []
+        if (
+            payload.get("iss") != Context.oauth_issuer
+            or Context.oauth_session_audience_id not in payload.get("aud", [])
         ):
             return False
         return True
--- a/backend/src/cms_backend/db/books.py
+++ b/backend/src/cms_backend/db/books.py
@@ -171,22 +171,32 @@ def get_zim_urls_prod(session: OrmSession, zim_ids: list[UUID]) -> ZimUrlsSchema
         )
         .where(
             and_(
-                Book.id.in_(zim_ids),
+                # Get all books for the related titles so we can determine the latest
+                Book.title_id.in_(select(Book.title_id).where(Book.id.in_(zim_ids))),
                 Book.needs_processing.is_(False),
                 Book.has_error.is_(False),
                 Book.needs_file_operation.is_(False),
                 Book.location_kind == "prod",
             )
         )
-        .order_by(Title.id, Book.flavour, Book.created_at.desc())
+        .order_by(Title.id, Book.flavour, Book.date.desc(), Book.created_at.desc())
     )
 
     result = ZimUrlsSchema(urls={zim_id: [] for zim_id in zim_ids})
-
-    # Filter to keep only one view link for the latest book per title+flavour
-    # combination
     seen: set[tuple[str | None, str | None]] = set()
+
     for row in session.execute(stmt).all():
+        is_latest = False
+        key = (row.title_id, row.book_flavour)
+
+        if key not in seen:
+            seen.add(key)
+            is_latest = True
+
+        # Only provide URLs if the book is one of the specifically requested zim_ids
+        if row.book_id not in result.urls:
+            continue
+
         if row.download_base_url:
             result.urls[row.book_id].append(
                 ZimUrlSchema(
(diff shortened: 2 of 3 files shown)


benoit74 (commented):
What if two books have the exact same date? Quite a very extreme edge-case, but I'm sure it will happen at some point and I would prefer to not have to fix the CMS at that moment ^^ Is it already working or do we end-up again with two books with a view URL instead of a single one?

elfkuzco:
> What if two books have the exact same date? Quite a very extreme edge-case, but I'm sure it will happen at some point and I would prefer to not have to fix the CMS at that moment ^^ Is it already working or do we end-up again with two books with a view URL instead of a single one?


When two books belonging to the same title and flavour have the same date, there will be two rows that have the `is_latest` set to True but because we use a set to ensure we only have one title+flavour combination, only the latest one (books are sorted by `created_at` as a last resort) will have the view URL. Added a new test for that and it works

benoit74:
Indeed, thank you!

benoit74 (changes requested):
Sorry, I just realize now that my specs were a bit too vague:
- for books in staging, we still want all books to have both a view and a download URL
- problem was only for books in prod where only the latest of the given title/flavour is present in catalog.xml / has a view URL

All books in staging are present in staging catalog.xml so they all have a view URL.

I feel like your PR is also modifying the logic for books in staging. Or do I miss something?

benoit74:
I'm also not sure we need the `latest_dates_subq` subquery, getting all books and only sorting by date + created_at and then filtering in-memory is probably sufficient and easier to read.

benoit74 (approved):
LGTM, thank you ; I hope I did not missed yet another edge case ^^

PR #263, opened 2026-04-29
rename user to account

### LINKED ISSUE #242: Move from `user` to `account`
Just like on Zimfarm, we need to move from `user` to `account` at API and DB level, for same reasons + consistency across our solutions now ^^

elfkuzco:
## Rationale
This PR renames user to account across the codebase

## Changes
- create migration to rename table from user to account along with renames and recreating constraints
- rename route from `/users` to `/accounts`
- replace word `user` with `account` across all functions and files

This closes #242


elfkuzco:
It appears `black` is formatting a line and still it's greater than 88 characters whereas Ruff does the right thing. Upgraded black manually and ran with `black --line-length=88 src/api/token.py` and it still produces line with 90 characters long. @benoit74 , should we use ruff only?

benoit74:
> It appears black is formatting a line and still it's greater than 88 characters whereas Ruff does the right thing. Upgraded black manually and ran with black --line-length=88 src/api/token.py and it still produces line with 90 characters long. @benoit74 , should we use ruff only?

This is the same problem I've experienced on Zimfarm. Did you check if there is not simply something wrong in black usage? (maybe we are not using the right parameter anymore ...)

benoit74 (approved):
LGTM, thank you

PR #264, opened 2026-04-30
use common book table component to show list of books

### LINKED ISSUE #236: Emphasizing "Created" column from book list in title details is weird
The list of books in title details is current emphasizing a lot the "Created" column, both because it is the first column and the only one to contain a link.

Image

I wonder if we should exchange "Created" and "Date" column positions and move the link to the "Date" column.

I'm not sure about that design either because when we have multiple flavours in a single title, then the "Date" has multiple times the same value, so it could be confusing as well.

Or maybe we should display the book ID (but I don't really like it) or simply remove the link and make the full row clickable (but then it is a bit hard to guess one can click on the row to "dive" into the book.

### LINKED ISSUE #241: Use same display for list of books
We currently have two places in the UI where we display a list of books: once on the inbox books, and once in the title details.

https://cms.openzim.org/inbox?tab=books
Image

https://cms.openzim.org/title/100r.co_en_all
Image

Having used quite a lot the CMS on recent days with real data, it became obvious that we need to merge these two displays into a single way to present things with a common Vue.JS component. 

I prefer title details way of showing things (there are many more useful details shown) except for the "status" column which is superior in books inbox. The "Deletion date" is also missing on title details (while it is quite an important info in fact).

elfkuzco:
## Rationale
This PR refactors the book table component so it can be used to show the list of books both in the title detail and inbox views

Screenshot_20260430_063358
Screenshot_20260430_063349

## Changes
- use props `isServerSide` to determine whether it's a server side table or just normal data table
- refactor `BookStatus` component to hide location when book is deleted/to be deleted when it's status text says  the same thing
- remove Location column since `BookStatus` now shows the locations always
- shorten `Deletion Date` format
- Remove `Created At` column which contains link for visiting book and make rows clickable with hover and cursor pointer effect indicating the row can be clicked

This closes #241 
This closes #236

### DIFF
--- a/frontend/src/components/BookStatus.vue
+++ b/frontend/src/components/BookStatus.vue
@@ -1,89 +1,50 @@
 <template>
-  <v-tooltip v-if="iconOnly" location="bottom">
-    <template #activator="{ props: tooltipProps }">
-      <span v-bind="tooltipProps">
-        <v-icon
-          v-if="isErrored"
-          size="small"
-          color="error"
-          icon="mdi-alert-circle-outline"
-        ></v-icon>
-        <v-icon
-          v-else-if="isProcessing"
-          size="small"
-          color="grey"
-          icon="mdi-clock-outline"
-        ></v-icon>
-        <v-icon v-else-if="isDeleted" size="small" color="grey-darken-2" icon="mdi-delete"></v-icon>
-        <v-icon
-          v-else-if="isToBeDeleted"
-          size="small"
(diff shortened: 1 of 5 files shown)


benoit74 (approved):
LGTM, thank you. I'm a bit concerned about the ability for users to discover that lines can be clicked on touch screens (tablets, ...) but I see that Github notifications (for instance) have mostly the same design, so let's live with it for now.

PR #266, opened 2026-04-30
add account management

### LINKED ISSUE #243: Implement user management screens
We need same screens to manage users than on Zimfarm. 

In CMS world, users are all accounts except those with `zimfarm` role. 

`zimfarm` role must not be assignable from the UI since it is a purely technical account we do not need to really manage. 

Just like on Zimfarm , accounts `viewer` must be hidden by default.

And we need the same "Create user" flows.

elfkuzco:
## Rationale

This PR adds API endpoints for account management and corresponding views in the UI
Screenshot_20260430_115028
Screenshot_20260430_114839
Screenshot_20260430_114816
Screenshot_20260430_114811
Screenshot_20260430_114804

## Changes]
- add `display_name` column to `account` table
- add migration script to copy username to display name and make username nullable
- add API endpoints to perform CRUD operations on accounts
- add views for accounts in the UI

This closes #243


benoit74 (approved):
LGTM, thank you

PR #267, opened 2026-05-04
fix oauth name used in can_decode check for session decoder

elfkuzco:
## Changes
- check for `oauth` in AUTH_MODES instead  of `oauth-session`


benoit74 (approved):
LGTM

PR #271, opened 2026-05-06
fix retention rules to always keep latest version of two recent months

### LINKED ISSUE #268: Book retention rules are still a bit incorrect
See https://cms.openzim.org/title/tomlehrersongs.com_en_all

Image

The oldest book (2025-06-01) should be marked for deletion and then deleted. Not sure why it is not happening.

elfkuzco:
## Rationale
This PR fixes the retention rules by filtering books at the application level instead of the DB level. Consequently, since all prod books belonging to the title are loaded, we revise the algorithm as follows:
- keep a list of the latest version from the two most recent periods. The two most recent periods may or may not include the recent/last month. This way, if there are books for the current month, then, they will surely take up the first slot. Older books may take up zero to two slots.
- for all books originally selected, if book is older than 30 days and book doesn't appear in the list built up, mark it for deletion

This fixes #268

### DIFF
--- a/backend/src/cms_backend/db/title.py
+++ b/backend/src/cms_backend/db/title.py
@@ -1,11 +1,11 @@
 import datetime
 from collections import defaultdict
 from pathlib import Path
+from typing import cast
 from uuid import UUID
 
 from psycopg.errors import UniqueViolation
-from sqlalchemy import Date, select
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
This is indeed the good algorithm and more intelligible than my explanations ^^

PR #273, opened 2026-05-07
differentiate between staging and prod books status

### LINKED ISSUE #234: Better indicate when books are in staging
It is very hard to differentiate book in staging from book in production in title detail books list and book detail views

Both use the verb "Published" and a green mark.

We should allow to better visually differentiate them:
- change icon for staging book (not sure which one is appropriate, but needs to indicate file is under inspection) and color (probably orange warning color is correct)
- replace "Published prod" and "Published staging" with "Published" (=prod) and "Staging" (=staging)

elfkuzco:
## Changes
- use eye icon for books in staging that don't have any errors
- chnage text from published to staging for books in staging that don't have any errors

Screenshot_20260507_092413

This closes #234


benoit74 (approved):
LGTM

PR #274, opened 2026-05-07
streamline shuttle move/copy operations

### LINKED ISSUE #233: Since we cannot really move file, rationalize codebase
Following https://github.com/openzim/cms/pull/232, we can probably simplify shuttle codebase since we always make a copy anyway, even when we simply want to move the file.

elfkuzco:
## Rationale
This PR simplifies the file movement operation of the shuttle. It does this in the following way:
- keep one current location as the source of truth since all the book data (file) is the  same across all current locations
- create the target locations using this first current location while simultaneously deleting any `current` location that isn't the original we are holding a reference to. Caveat is that book events are now showing as copies and not really as moves from one current location to another
- delete the original location and any other `current` locations that are left (especially if current location is greater than target locations)

This fixes #233

### DIFF
--- a/backend/src/cms_backend/shuttle/move_files.py
+++ b/backend/src/cms_backend/shuttle/move_files.py
@@ -70,71 +70,73 @@ def move_book_files(session: OrmSession, book: Book):
         book.needs_file_operation = False
         return
 
-    # start with copies
-    while len(target_locations) > len(current_locations):
-        current_location = current_locations[0]
-        target_location = target_locations[0]
+    # Grab one valid source file to copy from since it's the same
+    # file spread across multiple locations with the same book
+    # data
+    source_location = current_locations[0]
 
-        current_path = current_location.full_local_path(
-            ShuttleContext.local_warehouse_paths
-        )
+    current_locations_map = {
+        (loc.warehouse_id, loc.path): loc for loc in current_locations
+    }
+
+    for target_location in target_locations:
         target_path = target_location.full_local_path(
             ShuttleContext.local_warehouse_paths
         )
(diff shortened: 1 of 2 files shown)


benoit74 (changes requested):
- I realize it looks like we do not correctly handle the situation where a target location is same as a current one. Typical when we add a title to a second collection, we move from one current to the same target + another target. Or is this handled differently in terms of current/target locations?
- `Deleted old location`  => `Deleted book from`

elfkuzco:
> I realize it looks like we do not correctly handle the situation where a target location is same as a current one. Typical when we add a title to a second collection, we move from one current to the same target + another target. Or is this handled differently in terms of current/target locations?

Nope, you are correct.

backend/src/cms_backend/shuttle/move_files.py
```diff
@@ -100,9 +119,9 @@ def move_book_files(session: OrmSession, book: Book):
                 ShuttleContext.local_warehouse_paths
             )
             del_path.unlink(missing_ok=True)
-            logger.debug(f"Deleted old location for book {book.id} at {del_path}")
+            logger.debug(f"Deleted book {book.id} from {del_path}")
             book.events.append(
-                f"{getnow()}: deleted old location {loc_to_delete.full_str}"
+                f"{getnow()}: deleted book {book.id} from {loc_to_delete.full_str}"
```
benoit74: Do not put the book.id in book events, it is already a property of the DB row / object

benoit74 (changes requested):

benoit74 (approved):
LGTM

PR #280, opened 2026-05-12
clear collection titles before updating

### LINKED ISSUE #277: Impossible to add a second collection
When I try to add a second collection to an existing title, it fails.

For instance if I try to add "Endless / endless" to "academia.stackexchange.com_en_all", it fails. Looks like a problem with book location unique constraint, but I don't get why.

elfkuzco:
## Rationale
This PR clears the collection titles  before creating the new ones to be used for an update. By clearing the old collection titles in memory, we avoid creating duplicate collection titles when the original collection title is sent in the update reqeust. This is because SQLAlchemy list of collection titles isn't cleared with just `session.delete`. 
As a consequence, having duplicate collection titles leads to creating duplicate book locations which violates location unique constraint.

## Changes
- clear collection titles before adding new ones
- update tests to model real-world scenario such that title to update has book, collection title, etc.
- remove redundant second test since updated test covers everything

This fixes #277

### DIFF
--- a/backend/src/cms_backend/db/title.py
+++ b/backend/src/cms_backend/db/title.py
@@ -271,6 +271,8 @@ def update_title(
             for tc in title.collections:
                 session.delete(tc)
 
+            title.collections.clear()
+
             for entry in collection_titles:
                 collection = get_collection_by_name(
                     session, collection_name=entry.collection_name
--- a/backend/tests/db/test_title.py
+++ b/backend/tests/db/test_title.py
@@ -4,7 +4,14 @@ import pytest
 from sqlalchemy import select
 from sqlalchemy.orm import Session as OrmSession
 
-from cms_backend.db.models import Book, BookLocation, Collection, Event, Title
+from cms_backend.db.models import (
+    Book,
+    BookLocation,
+    Collection,
+    CollectionTitle,
+    Event,
+    Title,
+)
 from cms_backend.db.title import get_title_by_name_or_none, get_titles, update_title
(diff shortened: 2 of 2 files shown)


elfkuzco:
@benoit74 , I am wondering if this is not a good time to add a unique constraint to the collectiontitle table too. The keys for the constraint will be title_id, collection_id, and path. WDYT?

benoit74:
> @benoit74 , I am wondering if this is not a good time to add a unique constraint to the collectiontitle table too. The keys for the constraint will be title_id, collection_id, and path. WDYT?

We already have a unique constraint (primary key actually) on `title_id, collection_id` for `collection_title` table, since this is more restrictive I don't think we need another one. Do I miss something?

benoit74 (approved):
LGTM

elfkuzco:
> We already have a unique constraint (primary key actually) on title_id, collection_id for collection_title table,

No, we don't and we shouldn't because it acts as a many-to-many intermediary table for collection and title. But having a unique constraint made of `title_id`, `collection_id` and `path` will ensure we can't have the same title collection combination existing twice (at the DB level) although at the filesystem level, we intelligently avoid making duplicate copies on the same path.

benoit74:
But since `title_id, collection_id` is already unique, we are sure `title_id, collection_id, path` is unique as well. I don't get your point

elfkuzco:
> But since title_id, collection_id is already unique, we are sure title_id, collection_id, path is unique as well. I don't get your point

Yup, indeed. Just realized they are both used as primary keys. Sorry, my bad.

PR #285, opened 2026-05-19
add functionality to archive titles

### LINKED ISSUE #278: Archive/restore a title
How do we handle the situation of a title we do not want to publish anymore? 

Publisher can already delete all the books manually from the UI. This will only delete the ZIM file from storage and remove the book from catalog, but book object will still be there in "deleted" state for historical reason.

This allows to achieve main goal which is to not have this title in the catalog(s) and in the filesystem.

But the title is still there in CMS. Is this ok?

Do we need at least a more convenient button at title level to delete all books at once (which only remove ZIM files again)?

Do we need a way to "delete" the title itself, moving it to a different status where it is not shown in search results, not usable by new incoming ZIMs without publisher action, ...? 

Do we want to completely wipe it from DB to avoid keeping old artifacts forever? With a specific role e.g. only by an admin?

elfkuzco:
## Rationale
Screenshot_20260519_074307
Screenshot_20260519_074257
Screenshot_20260519_073859
Screenshot_20260519_073848
Screenshot_20260519_073840
Screenshot_20260519_073828
Screenshot_20260519_073809
Screenshot_20260519_103707
Screenshot_20260519_103659

## Changes
- add new permission for viewing archived titles
- prevent users without permission to view archived titles
- prevent books belonging to archived titles from being moved or recovered as archiving a title automatically marks the books for deletion. Similarly, restoring a title attempts to recover the books which have been marked for deletion. These operations undo any individual actions applied to books belonging to the title
- prevent attaching new books to archived titles
- refactor views to list titles into shared component for normal titles and archived titles. Archived titles allow selections to be made on them allowing user to perform bulk restore
- move logic to apply retention rules to seperate file as it causes cyclic imports between title and book modules which now share code

This closes #278


backend/src/cms_backend/db/title.py
```diff
@@ -131,9 +141,11 @@ def get_titles(
             Title.id.label("title_id"),
             Title.name.label("title_name"),
             Title.maturity.label("title_maturity"),
+            Title.archived.label("titl_archived"),
```
benoit74: titl_archived => title_archived (how could tests pass with this typo? Do we miss tests or is this unused?)

backend/src/cms_backend/db/title.py
```diff
@@ -345,87 +362,85 @@ def update_title(
...
-    return sorted(
-        books,
-        key=sort_fn,
-        reverse=True,
-    )
-
+    title.archived = True
+    now = getnow()
+    title.events.append(f"{now}: marked title as archived.")
+    logger.info("marking books belonging to title for deletion.")
```
benoit74: This log is not exactly useful since we miss details about title concerned. Either this is a debug log (and we consider we will have API call log near it to know which title is concerned) or we need to add title ID to the log

backend/src/cms_backend/db/title.py
```diff
@@ -345,87 +362,85 @@ def update_title(
...
 
-        for period in sorted_periods[:2]:
-            sorted_books_by_period = sort_books_by_filename_period(
-                books_by_period[period]
-            )
-            books_to_keep.add(sorted_books_by_period[0].id)
+    title.archived = False
+    now = getnow()
+    title.events.append(f"{now}: restored book from archive")
+    logger.info(
```
benoit74, line 427: Same remark

benoit74 (changes requested):
Thank you!

Shouldn't we add support for bulk archive operation as well? I'm sure it will help at some point, and it is a bit weird to support bulk restore but not bulk archive.

We can do it in another PR if it is simpler.

See other inline comments below.

backend/src/cms_backend/db/title.py
elfkuzco: tuple unpacking gave it a new name. fixing

elfkuzco:
> Shouldn't we add support for bulk archive operation as well? I'm sure it will help at some point, and it is a bit weird to support bulk restore but not bulk archive.


Added bulk archive for selected titles and updated list of images in PR description

benoit74 (approved):
LGTM, thank you

PR #288, opened 2026-05-19
add collection title path to xml path

### LINKED ISSUE #282: Optionaly include path attribute in catalog
`kiwix-serve` in `--catalogOnly` mode needs every book to have a `path` attribute as that's what it uses to build path prefixes.
FYI it is not using ZIM metadata (`Name`) for that because of conflicts that could occur.

In the non-CMS world, we'd generate a `path` attribute based on `url` attribute when the XML is to be used in `--catalogOnly`.

Now, I had to replicate that hack but it seems to me that this is something that the CMS should do, upon request.

Can we imagine passing a param (`?addPath` to /collections/xxx/catalog.xml`) and receive the catalog with `path` attr on books? Value would be the folder and the filename.

elfkuzco:
## Changes
- use collection title path while building `path` property for XML content of book

This closes #282

### DIFF
--- a/backend/src/cms_backend/api/routes/utils.py
+++ b/backend/src/cms_backend/api/routes/utils.py
@@ -55,7 +55,7 @@ def build_library_xml(
             if path_prefix.endswith("/"):
                 path_prefix = path_prefix[:-1]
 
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
👍

PR #291, opened 2026-05-22
omit deleted books in needs_attention flag

### LINKED ISSUE #290: Deleted books without titles must not be displayed in book inbox
Book inbox shows books without title since we assume we wanna assign them a title.

This is true "in-general" but wrong once we've decided to delete them and they have been deleted. The fact they are shown while "to-be-deleted" is not a problem because we want to see to-be-delete books anyway.

We currently have an occurrence in production:

Image

elfkuzco:
## Changes
- omit deleted books from condition to select books without title as those that need attention

This fixes #290

### DIFF
--- a/backend/src/cms_backend/db/books.py
+++ b/backend/src/cms_backend/db/books.py
@@ -82,7 +82,7 @@ def get_books(
         stmt = stmt.where(
             or_(
                 Book.location_kind.in_(["quarantine", "staging"]),
-                Book.title_id.is_(None),
+                and_(Book.title_id.is_(None), Book.location_kind.not_in(["deleted"])),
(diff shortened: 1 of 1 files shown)


elfkuzco:
@benoit74 , the API to retrieve books has quite man flags but the UI is hardcoded to query for books that "need attention" in the inbox view. I don't know if there are other flags we might want to apply this to or is this just sufficient for the UI. For example, there is a flag for `has_title` which can be used standalone (of course not by the UI becuaus it hardcodes `needs_attention` to True)

benoit74 (approved):
LGTM. I don't think we should alter other API filters.

PR #298, opened 2026-05-25
add metadata to title

### LINKED ISSUE #292: Handle more metadata at title level
While discussing #88 with @kelson42, we came to the conclusion that we need to handle way more metadata at CMS title level: in addition to the `Name` (which is very important when it comes to matching books with titles), a title must have many metadata. Some mandatory: `Title` (sic), `Creator`, `Publisher`, `Description`, `Language`, `Illustration_48x48@1`. Some optional (can be blank) : `LongDescription`, `License`, `Relation`, `Source`.

This means mostly all standard ZIM metadata besides `Flavour` (a title has multiple flavours), `Tags` (tags vary from book to book, e.g. `_pictures:yes/no`, so we keep this aside for now) and dynamic values like `Date`, `Counter`, `Scraper` (which we do not want to alter but rather keep as-is). Non-mandatory illustrations are also not handled at title level for the time being, future on this is not clear enough at this stage.

All books in a single title will share the same metadata values for metadata present at title level.

Whenever a book arrives in the CMS, there is two different situations.

First, if title has no book yet, then title metadata are automatically populated from book metadata.

Second, if title already has at least one book, then its metadata are already populated (by a former book) and the CMS hence compares new book metadata with title metadata. When they match, book continues with the rest of its lifecycle. When they do not match, book is forced through staging area with a status indicating that metadata do not match.

User can decide to update title metadata. A convenient UI allowing to easily compare metadata between book and title is necessary, and it should allow to easily transfer book metadata to title, only by one or all metadata; book inbox UI should also clearly display there is a metadata issue - note that there will be many other checks to display once M4 is completed. User can then still move the book to production manually (no matter if he decide to override title metadata or not). 

User can edit title metadata directly from title edit (we should move this edit from current popup to a tab, like on Zimfarm recipe edition).

When generating the XML catalog, metadata from the title have precedence (higher priority) over metadata from the book (which are unused), even if "blanked" for non-mandatory ones. Only metadata not handled at title level come from the book.

@elfkuzco do you feel like this could be ready to implement?

elfkuzco:
## Rationale
This PR adds metadata information to CMS titles. This allows having a central store for common metadata for all books belonging to a title. 

## Changes
- add various columns for storing metadata information
- add `warnings` column for storing warning-level messages to books. These warnings do not prevent a book from being acted upon by shuttle or cms but they prevent the book from being moved to prod even if it's maturity is `stable`
- update code to create/update title with necessary metadata
- implement views to create/update title and to sync title metadata from a book's metadata
- use title metadata as source of truth (if it exists) while generating catalog XML

Screenshot_20260525_161510
Screenshot_20260528_155326
Screenshot_20260528_155317
Screenshot_20260528_155307

This closes #292


benoit74 (changes requested):
Thank you!

First of all, we do not want to ask for metadata when creating a title, they will be populated by first book "entering" the title.

I'm also not convinced by the book `warnings` property naming ; wondering if `issues` is not a better naming at this stage, but I'm not totally bought by this.

Finally, I don't think that current UI is correct. I find the fact that we have the capability to "sync" book metadata to title in the book details a bit disturbing. First the "sync" word is confusing because it is hard to guess in which direction the copy will happen. Second the fact this must be done on the book while we edit title metadata is confusing as well. I also don't think that we should group metadata together in a "Metadata" section (aside from Title name, even if it is a metadata). I don't have strong proposition on how to move this forward, but I feel like we do not yet have correct solution.

Unless I missed it, we also miss a maintenance script to populate all title metadata from latest existing book.

elfkuzco:
> I'm also not convinced by the book warnings property naming ; wondering if issues is not a better naming at this stage, but I'm not totally bought by this.

Alternative would be to have a boolean column for storing the different issues a book might have. Currently, we have `has_error` but that determines if shuttle or mill will act on book. And you mentioned that there will be many checks to display once M4 is completed

Will these checks have the same impact as `has_error` or will they just be for signalling only?

elfkuzco:
> Finally, I don't think that current UI is correct. I find the fact that we have the capability to "sync" book metadata to title in the book details a bit disturbing. First the "sync" word is confusing because it is hard to guess in which direction the copy will happen. Second the fact this must be done on the book while we edit title metadata is confusing as well. 


Alternative would be to move it to the title view but that would require selecting from the list of books available to know which one to update the metadata from. Is that better?

elfkuzco:
> I also don't think that we should group metadata together in a "Metadata" section (aside from Title name, even if it is a metadata). I don't have strong proposition on how to move this forward, but I feel like we do not yet have correct solution.

Sorry, but I don't understand where you mean. There's no Metadata section in the title detail view in the UI.

benoit74:
> Will these checks have the same impact as has_error or will they just be for signalling only?

Most will be for signalling why the book has been forced to go through staging and needs manual validation to go to prod.

I think the fact we add a list column is correct, I'm only not totally convinced by the naming. Let's start with this for now and we will fix later if needed.

> Alternative would be to move it to the title view but that would require selecting from the list of books available to know which one to update the metadata from. Is that better?

Latest book always "wins" (no matter its flavour, all are expected to have same flavour).

> Sorry, but I don't understand where you mean. There's no Metadata section in the title detail view in the UI.

Let's forget about this, I tried to build a mockup below for title edit form (order of metadata is wrong, title must be after name and before description):

IMG_20260526_143418

- when metadata is identical to latest book, do not care to display anything
- when metadata is different, add a badge (probably not in read, orangish is probably better, like a warning) + show latest book value + a "Use this" button

Feel free to be only very vaguely inspired from this, it is more to show my points that anything to follow

elfkuzco:
updated PR description to show the new screenshot of the title edit tab which incorporates buttons to update from the latest book

benoit74 (commented):
I struggle to test this on my local dev instance. I would like to test the full flow from receiving a zimfarm notification to having the book in the catalog with different situations (title maturity is unstable, title maturity is stable but metadata differs, title is brand new and has not metadata, etc ...). I trust the automated tests to detect any regression we might introduce, but they are too hard to read to be sure we have implemented the right logic.

I used to rely on scripts in `dev` folder to create a meaningful configuration with the various use cases and wanted to see, but they are not maintained and cannot coexist ATM with a prod DB restored locally (same warehouse names, cannot easily wipe test objects while keeping the prod DB restore, ...).

How do you test these changes locally? Do you have custom tooling or do you simply patch DB to end-up in expected states for UI validation and 

I feel like it is important to improve this situation because we will only get more scenarii to test in the near future (bad number of articles in the ZIM, bad metadata, bad zimcheck output, ...). Maybe the script-based approach is not the right one and we should have something else, I don't know. But I'm sure I would like to not push code changes like this PR to production without having run some "end-to-end" tests (from zimfarm notification to catalog XML typically). We need to be aligned on this.

WDYT?

elfkuzco:
> How do you test these changes locally? Do you have custom tooling or do you simply patch DB to end-up in expected states for UI validation and

Yeah, I sometimes patch entries in the DB to end up in expected states for UI. Sometimes, I set up the tests so they can simulate a particular scenario. 

> I feel like it is important to improve this situation because we will only get more scenarii to test in the near future (bad number of articles in the ZIM, bad metadata, bad zimcheck output, ...). Maybe the script-based approach is not the right one and we should have something else, I don't know. But I'm sure I would like to not push code changes like this PR to production without having run some "end-to-end" tests (from zimfarm notification to catalog XML typically). We need to be aligned on this.

> WDYT?

I agree it's important but sometimes, I can only run the code without the shuttle because there's no file operations that would succeed. Currently, I have no ideas on how to go about it as it's a bit of a conundrum to me too. Maybe make the tests clearer as it's the only reliable way?

benoit74:
Good point regarding the shuttle. I did not understood yet why the shuttle wants to perform so many operations. Did you understood this? The shuttle was working fine without the CMS prod DB restored. Maybe we can enhance tooling to restore prod DB locally: we could have a script to create all expected files locally, they could probably be zero byte files, tbc

elfkuzco:
Updated dev scripts folder to create dummy files for any existing books in DB (to account for prod files). Files are created with `touch`. 
To distinguish between any entries in DB and dev content, dev objects start with prefix of `dev_`. This includes the warehouse names, collection names, title names, zimfarm notification filename and name metadata.

benoit74 (changes requested):
Very great, I like this!

Two minor tweaks needed before merge:
- you've added two files at backend/7389b26f-a00b-443c-a33d-b5bc9d26a955 and backend/d3e1078a-7ace-26bd-7eb3-5c8dd3a4c3f0 which were probably not supposed to be
- we should put more emphasis in TitleForm when title data differs from latest book data

I also wonder how/if we should clear book issues when they are solved (e.g. by updating title metadata). And how we could better show in book details when book metadata doesn't match title metadata. But let's keep this for separate issues I will open, it is far from clear to me how we wanna solve this.

I finally propose to add details about restoring a prod DB dump to `dev/README.md` (I'll share what I've drafted in a next comment).

benoit74:
<details><summary>dev/README.md new section</summary>
<p>


### Import production DB dump

If you have access to a production DB dump, you can restore it locally.

Mount you dump at `/data/cms` in PG container.

Drop and recreate the `cms` database:

```
docker exec -it cms_postgresdb bash -c \
  'psql -U cms -d postgres -c "DROP DATABASE cms WITH (FORCE);" -c "CREATE DATABASE cms;"'
```

Restore DB dump (assuming it is mounted in /data/cms)

```
docker exec -it cms_postgresdb bash -c \
  'pg_restore -U cms -d cms /data/cms'
```

Delete admin user so that it is recreated by API startup with admin/admin_pass credentials:

```
docker exec -it cms_postgresdb bash -c \
  "psql -U cms -d cms -c \"DELETE FROM account WHERE username='admin';\""
```

Restart the API:

```
docker restart cms_api
```

Create missing ZIM files locally so that shuttle operations works fine (it will create empty files with touch).

```sh
docker exec cms_mill python /scripts/setup_books.py
```


</p>
</details>

benoit74:
Nota: when I say more emphasis in TitleForm, I mean emphasis around each metadata which is different (like for instance the badge I proposed in my mockup, but could also just be an orange border, or say "Different from latest book which has" instead of "Latest book has" and put the text in orange, ...)

elfkuzco:
updated PR description with new title form showing emphasis

benoit74 (approved):
LGTM

Please rewrite commits before merging (probably squash in one or few commits), especially I don't want to have a merge from `main` to `feature` branch.

PR #299, opened 2026-05-26
view books and update book flavour

### LINKED ISSUE #293: View all books, filter them and edit book flavour
While openzim/cms#292 allows to edit many metadata, it doesn't cope with the fact that we sometimes need to also edit book `Flavour` (see openzim/zim-requests#1775). This can only be done at book level, and redone everytime a new book arrives.

We hence need to add the ability to edit book flavour (and only this metadata for now) from the UI.

Given openzim/zim-requests#1775 scenario, it seems like we will definitely need the ability to track down which book uses which flavour. We hence need a new "Books" tab (just like we have a "Titles" tab ATM), displaying a list of books with the ability to filter by name (free text) and/or flavour (v-select). 

@elfkuzco this is ready to implement IMHO. Do not hesitate to ask questions and suggest enhancements (e.g. in the list of book filters if you see something handy we should immediately include).

elfkuzco:
## Rationale
This PR enhances the UI to have a dedicated books view (different from the books tab in the Inbox view). Whereas the books in the inbox tab are set to fetch books that "need attention", the books view allows user to fetch even those that do not, along with newer filters. The PR also adds capability to update a book's flavour

Screenshot_20260526_120640
Screenshot_20260526_120620

## Changes
- create dedicated book view for listing all books
- add capability to edit book flavour via API/UI
- create different set of filters for the new book view differnt from the inbox view filters

This closes #293

### DIFF
--- a/backend/src/cms_backend/api/routes/books.py
+++ b/backend/src/cms_backend/api/routes/books.py
@@ -14,11 +14,16 @@ from cms_backend.db.book import delete_book as db_delete_book
 from cms_backend.db.book import get_book as db_get_book
 from cms_backend.db.book import move_book as db_move_book
 from cms_backend.db.book import recover_book as db_recover_book
+from cms_backend.db.book import update_book as db_update_book
+from cms_backend.db.books import get_book_flavours as db_get_book_flavours
 from cms_backend.db.books import get_book_languages as db_get_book_languages
 from cms_backend.db.books import get_books as db_get_books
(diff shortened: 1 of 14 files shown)


elfkuzco:
Not sure why CI isn't running checks...

benoit74:
You should try to amend commit (with nothing different, just a new commit ID) and push again. Probably a bug githubside.

benoit74 (approved):
Please check CI is green and then you can merge, LGTM

PR #307, opened 2026-05-28
set catalog flavour from book

### LINKED ISSUE #306: Catalog does not uses overridden book flavour
When we override the book flavour, this new value is taken into account to compute books to expose in the catalog XML but the XML itself seems to contain original book flavour instead of overridden one.

elfkuzco:
## Changes
- set catalog flavour from book instead of zim metadata since book flavours can now be updated (see #299)

This fixes #306


benoit74 (approved):
LGTM

PR #313, opened 2026-06-02
skip deleted and to_delete books from title modification events

### LINKED ISSUE #312: Book marked for move despite being deleted
Any idea how https://cms.openzim.org/book/570959af-5a5f-03ab-f1ad-21c23d299a5d ended up in current status? Looking at event log, the situation is quite weird.

@elfkuzco can you investigate based on event logs if you find an explanation how this could have happened? 

My understanding is:
- book was in quarantine pending title
- book has been marked for deletion and delete
- new book arrived with same name
- we create the title
- both books have been associated with newly created title
- both book have been marked for move despite oldest one been already deleted

The fact that both books are associated with newly created title is probably ok. Even if a bit weird for an already deleted title, it probably does not really harm and even allows to limit amount of orphaned books. But the deleted book should not have been updated for move, this is not relevant / possible.

elfkuzco:
## Changes
- omit books in `deleted` and `to_delete` location kinds from being processed in title modification events

This fixes #312

### DIFF
--- a/backend/src/cms_backend/mill/process_title_modifications.py
+++ b/backend/src/cms_backend/mill/process_title_modifications.py
@@ -42,6 +42,7 @@ def process_title_modifications(session: OrmSession):
                 Book.title_id.is_(None),
                 Book.has_error.is_(False),
                 Book.name == title_name,
+                Book.location_kind.not_in(["deleted", "to_delete"]),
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM

PR #315, opened 2026-06-03
update book issues

### LINKED ISSUE #310: Recompute book issues
https://github.com/openzim/cms/pull/298 introduced "issues" in books, and especially the `metadata mismatch` issue. https://github.com/openzim/cms/issues/294 introduced the `flavour mismatch` issue.

Currently, these `issues` list is however not updated as book or title configuration is changed. For instance if title metadata are updated with book ones, or if title flavours are updated so that book can make it to the title.

This is not blocking anything in terms of book processing, but it presents "wrong" information in the UI.

We need to enhance current logic so that the list of issues is recomputed. 

I think that for now we can probably recompute issues only when:
- book configuration is changed 
- title configuration is changed (recompute all books issues)

In the future, we will probably have other triggers, which will recompute book issues as well.

@elfkuzco this is probably ready for development

elfkuzco:
## Rationale
This PR adds a helper function to update a book's list of issues either during title update or book update. Book issues are only updated if:
- book has associated title
- book has necessary metadata and
- associated title has necessary metadata

This closes #310


benoit74 (changes requested):
Testing this, I realize just now that we do not want to update book issues of deleted or to_be_deleted books. It makes no sense to indicate that a book which is gone (or mostly gone) has issue(s). We still need to update issues when book is restored (from to_be_deleted).

benoit74 (approved):
LGTM

PR #318, opened 2026-06-04
add/remove language codes and refuse to publish books with invalid codes

### LINKED ISSUE #101: Refuse to publish books with unknown ISO639-3 codes in `Language` metadata
The CMS must refuse to publish books with unknown ISO639-3 codes in `Language` metadata.

`mul` special value must also be forbidden (even if part of the ISO639-3), since the specification says that we must list all languages used in this metadata.

This will obviously change over time as ISO639-3 norm gets updated. We should hence rely on a widely popular and regularly updated library / database for this check.

We might need to patch this library/database from time to time when it gets outdated as the norm evolves but the library/database is not yet updated.

This issue comes from https://github.com/openzim/overview/issues/51#issuecomment-2904587084 discussion

elfkuzco:
# Rationale
This PR enhances the mill to ensure that a book's language metadata is a valid ISO639-3 language code. Additionally, it provides ability to extend the list of language codes and also disallow some that might be valid

## Changes
- add env vars `DISALLOWED_LANGUAGE_CODES` and `CUSTOM_LANGUAGE_CODES` to support removal and adding of language codes respectively
- update language code database at mill startup
- mark books with unknown language codes as errored and refuse to add book to title for further processing

This closes #101

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "humanfriendly == 10.0",
     "Werkzeug == 3.1.5",
     "xxhash == 3.7.0",
+    "pycountry == 26.2.16",
 ]
 dynamic = ["version"]
 
--- a/backend/src/cms_backend/__init__.py
+++ b/backend/src/cms_backend/__init__.py
@@ -1,5 +1,7 @@
 import logging
 
+import pycountry
+
 from cms_backend.context import Context
 
 logger = logging.getLogger("backend")
@@ -9,3 +11,17 @@ if not logger.hasHandlers():
     handler = logging.StreamHandler()
     handler.setFormatter(logging.Formatter("[%(asctime)s: %(levelname)s] %(message)s"))
     logger.addHandler(handler)
+
+
+def update_language_codes():
+    for code in Context.disallowed_language_codes:
+        try:
+            pycountry.languages.remove_entry(alpha_3=code)  # pyright: ignore[reportUnknownMemberType]
+        except Exception as exc:
+            logger.warning(f"failed to remove language code '{code}': {exc}")
+
+    for code in Context.custom_language_codes:
+        pycountry.languages.add_entry(alpha_3=code)  # pyright: ignore[reportUnknownMemberType]
+
+
+update_language_codes()
--- a/backend/src/cms_backend/api/routes/fields.py
+++ b/backend/src/cms_backend/api/routes/fields.py
@@ -1,6 +1,7 @@
 import base64
 from typing import Annotated, Any
 
+import pycountry
 from pydantic import (
     AfterValidator,
     Field,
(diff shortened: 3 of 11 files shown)


elfkuzco:
fixing issues...

benoit74 (changes requested):
I'm sorry, I realize I've not been specific enough in the issue. 

This language metadata issue should not totally block books but simply force them in staging with a new book issue kind, just like other book issues we already have (metadata mismatch, flavour mismatch). Since language metadata in catalog endpoint now anyway comes from the title, this will be OK.

elfkuzco:
> Since language metadata in catalog endpoint now anyway comes from the title, this will be OK.

But the issues list won't be cleared because we don't update language individually for books which would trigger `update_book_issues`. Should we support updating languages for books?

benoit74:
> But the issues list won't be cleared because we don't update language individually for books which would trigger update_book_issues. Should we support updating languages for books?

Nope, either we set a correct language value in title (and book still has a bad language) or we update pycountry / env vars (and reprocess book to clear the issue).

elfkuzco:
had to rebase so i could get the features from the `update_book_issues`.

benoit74 (changes requested):
We are getting closer, but `Language` ZIM metadata can be a comma-separated list of languages, and it looks like this PR is not ready for that (each language in the list must be valid)

benoit74 (changes requested):
`validate_language_code` must support CSV of languages as well

elfkuzco:
> `validate_language_code` must support CSV of languages as well

currently, it's only used as the Title's language. Will a title's language be a CSV of languages? I don't use it for in book related schema as we don't update book languages yet.

benoit74:
Title language is under normal condition strictly identical to book language, just like any title metadata coming from the book. The title language can hence be a CSV as well

benoit74 (changes requested):
I now realize that this change means we will refuse to create or update titles with bad language metadata. Which is quite the good move, except that we should update UI just like in Zimfarm to:
- properly display 422 errors from the API (currently it looks like they are not displayed at all)
- ideally display error(s) early in the UI (as user type typically)

Sorry for clarifying this PR scope so late

elfkuzco:
> ideally display error(s) early in the UI (as user type typically)

not sure if this is feasible because the list of valid language codes are dynamic and we can't guarantee it is valid. only validator would be the lengths of each code are up to 3 characters after splitting by `,`

elfkuzco:
Screenshot_20260608_132452

benoit74:
> not sure if this is feasible because the list of valid language codes are dynamic and we can't guarantee it is valid. only validator would be the lengths of each code are up to 3 characters after splitting by ,

This is what I suspected, thanks for confirming ^^

benoit74 (approved):
LGTM

PR #319, opened 2026-06-04
dump update data in json mode

### LINKED ISSUE #316: Impossible to set collection download base URL from UI
I tried to set this collection download base URL from the UI but it failed with a bad error (it was blank before). Looks like it is yet a problem with AnyUrl and str conversion.

Image

elfkuzco:
## Changes
- dump pydantic models in `json` mode before using in update statement

This fixes #316


benoit74 (approved):
LGTM

PR #321, opened 2026-06-05
create title from book or add title to book

### LINKED ISSUE #311: Easily solve problem of book pending title
When a book is stuck in quarantine because it misses proper title, we should have a handy UX to solve the situation:
- either associate book with an existing title, even if title name does not match ; propose to update title name with book one (optionally) ; user should be able to search for title to choose
- or create a new title matching book metadata (name, etc ...)

UI could be to have a "Add to existing title" and "Create new title" action button in book details in such a situation, with proper popup.

elfkuzco:
## Rationale
This PR improves the UI to create a title from book details or associate a book with an existing title. This way, pending books can easily be associated with books as title modification/creation register events which lead to re-processing of pending books
Screenshot_20260605_072022
Screenshot_20260605_070807
Screenshot_20260605_065336
Screenshot_20260605_065321

## Changes
- make TitleTable reusable and optionally disable update of routes when table limit is changed or filters are updated. used in title selection dialog when updating tilte name with book name
- sync title table page with paginator and fix issues with reloading from page query parameter
- restrict title modification events from working on archived titles

This closes #311

### DIFF
--- a/backend/maint-scripts/update_titles_from_latest_books.py
+++ b/backend/maint-scripts/update_titles_from_latest_books.py
@@ -9,7 +9,8 @@ from sqlalchemy.orm import Session as OrmSession
 from cms_backend import logger
 from cms_backend.db import Session
 from cms_backend.db.models import Book, Title
-from cms_backend.db.title import get_title_by_id, title_is_missing_mandatory_metadata
+from cms_backend.db.rules import title_is_missing_mandatory_metadata
+from cms_backend.db.title import get_title_by_id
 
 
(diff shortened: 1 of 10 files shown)


benoit74 (approved):
LGTM, well done

elfkuzco:
when populating the language in the title form, i want to select the first language in the list in case it has metadata has multiple languages. wdyt?

https://github.com/openzim/cms/pull/318#issuecomment-4621719736

benoit74:
> when populating the language in the title form, i want to select the first language in the list in case it has metadata has multiple languages. wdyt?

Nope, title must have all languages just the the book (in a CSV)

PR #324, opened 2026-06-05
fix flavour mismatch comparison during book move

### LINKED ISSUE #323: Books without flavour cannot be moved to production
When I have a book in staging without any flavour, it is impossible to move it production. CMS complains that flavours do not match:

https://cms.openzim.org/book/3f55f47b-afd4-b841-ef31-71a3f8edd0c8

Image

elfkuzco:
## Changes
- refuse to move book to prod only if title has flavours and book flavour not in list of flavours

This fixes #323

### DIFF
--- a/backend/tests/db/test_book.py
+++ b/backend/tests/db/test_book.py
@@ -14,6 +14,7 @@ from cms_backend.db.book import (
 )
 from cms_backend.db.exceptions import RecordDoesNotExistError
 from cms_backend.db.models import Account, Book, Title, ZimfarmNotification
+from cms_backend.db.rules import has_flavour_mismatch
 from cms_backend.schemas.models import BookUpdateSchema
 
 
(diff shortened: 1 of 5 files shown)


benoit74 (changes requested):
We need to check the opposite: if book has a flavour but title has no flavour this should raise.

backend/tests/db/test_book.py
```diff
@@ -327,3 +328,19 @@ def test_update_book_flavour_mismatch_issues(
...
+
+
+@pytest.mark.parametrize(
+    "book_flavour, title_flavours, expected",
+    [
+        pytest.param(None, ["maxi"], True),
+        pytest.param("maxi", ["maxi", "mini"], False),
+        pytest.param("maxi", ["mini", "nopic"], True),
+        pytest.param(None, [], False),
+        pytest.param("maxi", [], True),
```
benoit74, line 340: Please ensure following cases do work as well (or add support for them):

- `pytest.param("", [], False),` 
- `pytest.param("", ["maxi"], True),`

These are not supposed to be found in DB at this stage, but I'm worried we might change that at some point and not realize it has consequence on this feature.

benoit74 (changes requested):

elfkuzco:
seems codecov is having issues uploading test reports

benoit74 (approved):
LGTM, please upgrade codecov CI action to `v7` in a distinct PR, we are still on `v5` and this is probably the issue. We can live without codecov upload for this PR.

PR #325, opened 2026-06-07
backup books

### LINKED ISSUE #297: Backup a book
There are some books of particular interest. Either because we need them to analyze a bug, or they showcase an important milestone, etc ... 

These backup books must continue to live no matter what happens to them due to the regular CMS lifecycle. We need for instance to backup a book which is in staging just like a book in production or even in quarantine.

Book needs hence to have a "backup" property which can be set from the UI ("Backup" button). When a book is backup, it is immediately copied to a new "Backup" area. Just like we already have "Quarantine" and "Staging", this area definition in terms of storage is done through specific env var. Backup also has a download + view URL to display in book details (just like any URL).

In the UI, when a user decides to delete a book the popup:
- informs when the book is already backup: "This book will continue to live in backup"
- propose to backup with a checkbox: "Preserve this book in backup"

Backup books are shown through a special list (once #293 is implemented) just like archived titles.

@elfkuzco I think this is ready to be implemented, do no hesitate to ask question / make suggestions

elfkuzco:
## Rationale
This PR enhances the API/UI to make backups  of books. This allows the "permanent" storage of book files even after they are deleted.
Screenshot_20260607_110036
Screenshot_20260606_233801
Screenshot_20260606_232650
Screenshot_20260606_233744
Screenshot_20260606_233434

## Changes
- add endpoint and db functions to backup a book
- backup book first and create an event to delete the book later if user wants to delete a book but preserve it first
- add `is_backup` property to book locations to determine if the current location is a backup location and if book has backup
- omit backup locations from `get_zim_urls_prod` and `get_zim_urls_staging`
- add `get_zim_urls_backup` for buliding up urls of backup locations
- omit backup book locations from catalog.xml output (both for prod and staging) since joining with book locations can create duplicate book entries in query to get books.
- add background task to process deletion events scheduled when user wants to create a backup and delete later
- rename env `PROCESS_TITLE_MODIFICATIONS_INTERVAL` to `PROCESS_EVENTS_INTERVAL` as there might be more events
- omit backup locations during delete files action in shuttle
- omit backup locations during move operation in shuttle
- avoid deleting original source locations during a backup operation
- refactor `get_missing_keys` to only flag a value as missing if it is `None` or `""` . This way, when book deletion event is created with `force_delete` set to `False`, the `get_missing_keys` function doesn't see the `False` value as a missing value
- update dev README on new warehouse and hazards of mingling operations with prod DB entries and those created by setup scripts
- add view to see books that are currently backed up (based on current filters)

This closes #297


elfkuzco:
@benoit74 , CI is failing due to coverage being unable to upload results and doesn't have to do with tests

backend/src/cms_backend/db/book.py
```diff
@@ -557,3 +577,68 @@ def update_book_issues(session: OrmSession, book: Book, *, update_events: bool =
...
+        (loc for loc in book.locations if loc.status == "current" and loc.is_backup),
+        None,
+    )
+
+    if existing_backup:
+        raise ValueError("Book already has a backup.")
+
+    # Create a current and tagret location for the backup using the book's details
+    target_locations = [
+        FileLocation(
```
benoit74, line 639: This is wrong, target locations is expected to be the list of all locations where the ZIM must be placed. It must hence be all current locations + the backup location. I believe it would allow to get rid of the complexities you had to introduce in the shuttle move files.

backend/src/cms_backend/shuttle/move_files.py
```diff
@@ -48,8 +48,10 @@ def move_book_files(session: OrmSession, book: Book):
         )
         return
 
+    # exclude backup locations from the list of current locations as we
```
benoit74, line 51: I don't think you need any change to the move_files operation if target_locations is set appropriately as "all locations where we expect the file to be after move_files operation" (see other comment)

benoit74 (changes requested):
I feel like supporting backup + delete in a single operation is adding too much complexity. It would have been convenient, but looking at the codebase it does not seems to be right to me. It also makes things way harder to understand because some delete are done immediately while others don't, plus we end-up with a book in a very weird transient state after backup is requested and before deletion event queued and processed. Book events says nothing about the pending deletion and it looks like the deletion has been forgotten.

It also makes the "backup" feature way too prominent (visible in all deletes) while it should be a niche case for very rare situations.

Finally, I failed to make it work, it failed on my local instance both for a book in staging and a book in production.

I propose we rather consider it is impossible to backup + delete at the same time. One will have to first backup, and then delete.

elfkuzco:
> Finally, I failed to make it work, it failed on my local instance both for a book in staging and a book in production

What was the issue?

> I propose we rather consider it is impossible to backup + delete at the same time. One will have to first backup, and then delete.

Okay.

backend/src/cms_backend/db/book.py
elfkuzco: if the target location is filled with the current locations plus the backup location, the  book events will be filled with:
```py
            book.events.append(
                f"{getnow()}: left book at identical location "
                f"{target_location.full_str}"
            )
```

is that desirable?

elfkuzco: unless we remove the event and just log only

backend/src/cms_backend/shuttle/move_files.py
elfkuzco: true

elfkuzco: on re-evaluation, during normal book move, we can create new target locations (maybe to staging) and this doesn't include the backup location in it's targets. so, when we do the move, we will delete the backup because it's in the remaining current locations after we have copied the books over to the target (which doesn't include the backup)

elfkuzco:
> Finally, I failed to make it work, it failed on my local instance both for a book in staging and a book in production.

I think maybe you need to wipe the dev entries and recreate. This could be due to because the backup warehouse not being created yet in your db. so, you probably want to re-run the setup scripts

frontend/src/components/BooksBaseView.vue
```diff
@@ -0,0 +1,321 @@
...
+import { useAuthStore } from '@/stores/auth'
+import type { BookLight } from '@/types/book'
+import type { Paginator } from '@/types/base'
+import { computed, onBeforeUnmount, onMounted, ref, watch } from 'vue'
+import { useRouter, useRoute } from 'vue-router'
+
+interface Props {
+  routeName: string
+  hasBackup?: boolean
+  showBackupFilter?: boolean
```
benoit74: This `showBackupFilter` is unused / unimplemented. I'm not really sure what you wanna show/hide with this filter.

frontend/src/components/BooksBaseView.vue
```diff
@@ -0,0 +1,321 @@
...
+})
+const errors = ref<string[]>([])
+
+const bookFilters = computed(() => {
+  const query = router.currentRoute.value.query
+  const derived = {
+    id: '',
+    name: '',
+    flavour: '',
+    needs_attention: 'no',
```
benoit74, line 100: This default value makes no sense when in backup view, we should at least set it to "All" or even completely drop this filter.

backend/src/cms_backend/db/books.py
```diff
@@ -301,11 +311,81 @@ def get_zim_urls_staging(session: OrmSession, zim_ids: list[UUID]) -> ZimUrlsSch
...
+                Book.needs_file_operation.is_(False),
+                BookLocation.is_backup.is_(True),
+            )
+        )
+        .order_by(Title.id, Book.flavour, Book.date.desc(), Book.created_at.desc())
+    )
+
+    result = ZimUrlsSchema(urls={zim_id: [] for zim_id in zim_ids})
+
+    for row in session.execute(stmt).all():
```
benoit74, line 352: Both backup_download_base_url and/or backup_view_base_url could be blank, in which case we do not add corresponding URL

benoit74 (changes requested):
Works great now. And no, I did add the warehouse in DB, at least now I do not have anymore problem in my manual tests.

benoit74 (approved):
LGTM

PR #330, opened 2026-06-08
add task to automatically mark staging books for deletion

### LINKED ISSUE #44: Files posted in .hidden/dev should have an expiry date
From zimfarm created by [Popolechien](https://github.com/Popolechien): openzim/zimfarm#650  

(not sure if this is more appropriate in /CMS, but anyway)

Files that are assigned to .hidden/dev are mostly to test out whether a given recipe works and their long-term value is negligible. They linger and accumulate in dev.library.org, which itself has quite a few limitations in the coherence of things it displays (it could for example list files by order of latest creation).

I therefore don't think there would be any visible issue if files were deleted after a month (or even 10 days).

elfkuzco:
## Changes
- add background task to mark staging books for deletion
- change order clause of `get_books` function to prioritize books in `staging` and `to_delete` when `needs_attention` is True
- add new env vars for task to delete staging books created before a specific time

This closes #44

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -1,3 +1,4 @@
+import datetime
 from typing import Any, Literal
 from uuid import UUID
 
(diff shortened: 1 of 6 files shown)


benoit74 (approved):
LGTM ; I realize that book ordering is still probably not perfect, but I don't have something clear in mind so let's merge with this order for now, it will be easy to fix anyway.

PR #332, opened 2026-06-09
remove book backup

### LINKED ISSUE #328: remove a backed up copy of a book
While #297 introduce the possibility to backup a book, we need to be able to "un-backup" it, i.e. remove the "backup" mark and the backup location. It is important to support this scenario both when book is "deleted" and when book is under usual statuses ("quarantine", "staging", "prod").

I'm not really convinced by "unbackup" term. And this must not be confused with #327.

I think this is ready to be developed.

elfkuzco:
## Rationale
This PR introduces a way to remove existing book backups by resetting the existing backups `is_backup` property to False and creating new target locations. This way, the `move_files` function deletes the extra current locations.

Screenshot_20260609_100117
Screenshot_20260609_100108
Screenshot_20260609_100056
Screenshot_20260609_100049

## Changes
- scope `is_backup` property to indivdiual `FileLocation` object instead of function to create book target locations as that applies it to all
- add function to remove book backup

This closes #328


benoit74 (approved):
LGTM

PR #335, opened 2026-06-09
reuse existing current locations in backup operation

### LINKED ISSUE #334: book backup moves file from staging folder to collection foldere
Image

In the above image, book was originally in `dev_hidden/staging` but backup logic moves it to `dev_prod/wikipedia` because of the way target locations are built up using the collection path instead of existing locations information
https://github.com/openzim/cms/blob/9f656fa61f8b794dc8ebc86352a5599917584e82/backend/src/cms_backend/db/book.py#L634-L640

elfkuzco:
## Changes
- reuse existing current locations to build up list of target locations during backup instead of collection titles path

This fixes #334

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -631,11 +631,10 @@ def backup_book(
     if existing_backup:
         raise ValueError("Book already has a backup.")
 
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM, nice catch

PR #336, opened 2026-06-09
Recover deleted book with backup

### LINKED ISSUE #327: Recover a deleted and backup book
When a book "Deleted" and "Backup", we need to be able to recover the book back into the normal lifecycle. It means that we will have to "re-inject" the book back to full processing, like if it arrived from a zimfarm notification. Except that if might already be associated with a title for instance, but we need to check again for book issues, etc ...

This is ready to be implemented

elfkuzco:
## Changes
- update a matching location during `move_files` using the properties of it's target
- create helper function to recover deleted files that have a backup

This closes #327

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -360,15 +360,24 @@ def recover_book(session: OrmSession, book_id: UUID) -> Book:
     book = get_book_or_none(
         session,
         book_id,
-        needs_file_operation=True,
         needs_processing=False,
-        locations=["to_delete"],
+        locations=["to_delete", "deleted"],
     )
-    if book is None or (book.deletion_date and book.deletion_date <= now):
+
+    if book is None:
         raise RecordDoesNotExistError(f"Book {book_id} is not eligible for recovery.")
 
     if book.title and book.title.archived:
         raise ValueError(f"Book title {book.title_id} is currently archived")
+
+    if book.location_kind == "deleted" and book.needs_file_operation is False:
+        return _recover_deleted_book(session, book)
+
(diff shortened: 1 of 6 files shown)


backend/src/cms_backend/db/book.py
```diff
@@ -721,3 +729,68 @@ def remove_book_backup(
...
+    )
+
+    if not existing_backup:
+        raise ValueError("Book does not have a backup.")
+
+    target_filename = existing_backup.filename
+    # update the book location so we can update book issues
+    book.location_kind = "staging"
+    update_book_issues(session, book)
+    goes_to_staging = book.title.maturity != "stable" or len(book.issues) != 0
```
benoit74: We should move this formula `book.title.maturity != "stable" or len(book.issues) != 0` to a shared location so that we have same logic in both places using it.

backend/src/cms_backend/db/book.py
```diff
@@ -721,3 +729,68 @@ def remove_book_backup(
...
+                target_filename,
+            )
+        ]
+        if goes_to_staging
+        else [
+            FileLocation(tc.collection.warehouse_id, tc.path, target_filename)
+            for tc in book.title.collections
+        ]
+    )
+    # move_files does not consider backup locatons as sources as it removes excessive
```
benoit74: Nice catch !

`backup locatons` => `backup locations`

benoit74 (changes requested):
Small suggestions

elfkuzco:
resolved conflicts as branch was originally based on intermediate branches that have now been closed.

benoit74 (approved):
LGTM

PR #337, opened 2026-06-09
define collection item count thresholds

### LINKED ISSUE #212: Detect significant changes in same title ZIM files
If a new (staging) ZIM has significant changes (for example, very larger or very smaller file size) in comparison to the previous published version (of the dame title/flavour), then it should trigger a warning and it should be manually validated (before publishing).

What "significant changes" exactly means has obviously to be defined and will be a heuristical tradeoff.

From a user perspective, the CMS behaviour should be very similar to the scenario when the zimcheck quality tradeoffs are not reached.

This is necessary to avoid to publish ZIM files which are broken, but not in a way that the scraper or zimcheck woukd be able to detect it.

elfkuzco:
## Changes
- create new columns for recording collection defined article and media count change thresholds
- update book issues if book media/article count differs from the latest book count
- compute minimum threshold from collections to which book title belongs to
- adapt API/UI to receive new values when creating/updating collection
- 
Screenshot_20260609_182722
Screenshot_20260609_182708
Screenshot_20260609_182658
Screenshot_20260609_182642

This closes #212


backend/src/cms_backend/db/book.py
```diff
@@ -595,8 +595,66 @@ def update_book_issues(session: OrmSession, book: Book, *, update_events: bool =
...
+    if latest_book is None:
+        latest_book = book
+
+    collection_media_count_change_threshold = min(
+        [
+            tc.collection.media_count_change_threshold
+            for tc in book.title.collections
+            if tc.collection.media_count_change_threshold
+        ],
+        default=1.0,  # 100% difference as default
```
benoit74: Please allow to customize defaults with an environment variable which default if unset is rather 10% (0.1).

backend/src/cms_backend/db/book.py
```diff
@@ -595,8 +595,66 @@ def update_book_issues(session: OrmSession, book: Book, *, update_events: bool =
...
+        .where(
+            Book.location_kind == "prod",
+            Book.id != book.id,
+            Book.has_error.is_(False),
+            Book.needs_processing.is_(False),
+            Book.needs_file_operation.is_(False),
+            Book.title_id == book.title_id,
+            Book.flavour == book.flavour,
+            BookLocation.status == "current",
+            BookLocation.is_backup.is_(False),
```
benoit74: Why should we care about book location?

backend/src/cms_backend/db/book.py
```diff
@@ -595,8 +595,66 @@ def update_book_issues(session: OrmSession, book: Book, *, update_events: bool =
...
             )
 
+    # Get the latest prod book that isn't the current book
+    latest_book = session.scalars(
+        select(Book)
+        .join(BookLocation)
+        .where(
+            Book.location_kind == "prod",
+            Book.id != book.id,
+            Book.has_error.is_(False),
```
benoit74: Why should we care about book status (error, needs_processing, needs_file_operation ...)? I think this should not influence the fact it is the latest book.

frontend/src/components/CollectionForm.vue
```diff
@@ -49,6 +49,30 @@
...
+
+    <v-text-field
+      v-model="formData.article_count_change_threshold"
+      label="Article Count Change Threshold (Optional)"
+      type="number"
+      min="0"
+      max="1.0"
+      density="comfortable"
+      variant="outlined"
+      placeholder="0.1"
```
benoit74, line 61: Shall we rather use the actual default value when not set at collection level (the one coming from context / environment variable)?

Probably means that we need a new API to give this to the UI, but I'm not against exposing these configuration details to the "world" since it might help. Not all context should be exposed, but important knobs like this one make sense to be visible.

benoit74 (changes requested):

backend/src/cms_backend/db/book.py
elfkuzco: tried to mimick the latest book which we show in the catalog endpoint

backend/src/cms_backend/db/book.py
elfkuzco: okay

backend/src/cms_backend/db/book.py
benoit74: Yeah, but I would not expect a production book to not have a current location, or DB is corrupted anyway and we have a serious issue.

elfkuzco:
added a new endpoint to retrieve api config values and merge with the one defined by `config.json` (this takes precedence). should we add more values in the `/config` endpoint?

benoit74 (approved):
LGTM, thank you

benoit74 (changes requested):
Sorry, just saw some tests are failing now. Please fix them

benoit74 (approved):
LGTM

PR #338, opened 2026-06-11
merge titles

### LINKED ISSUE #279: Merge two titles
Typically when a misconfiguration is done, we end-up in a situation where we have two titles instead of one.

This is for instance currently the case for "radiozamaneh" titles:

https://cms.openzim.org/titles?name=radiozamaneh

Image

We have two titles where in fact there is (should be) only one, it is just that ZIM name (or title) has been fixed from `radiozamaneh-com_far_all` to `radiozamaneh-com_fas_all` in Zimfarm. Here this duplicate has been badly created by the CMS population script we ran few weeks ago, because the ZIMs did existed on the file system (where they should not).

But anyway, we can end-up again in such a situation in the future, where a publisher create a new title by mistake when an existing title should have been renamed in fact.

The "natural" way to solve this is to move books from "bad" `radiozamaneh-com_far_all` title to "good" `radiozamaneh-com_fas_all`, delete the empty title and then let retention rules clean the oldest ones if needed.

Is this a new action or only a variation of title deletion expressed in https://github.com/openzim/cms/issues/278

elfkuzco:
## Rationale
This PR adds the ability to merge multiple titles into a single title. As a consequence, books from the source titles are associated with the new title. Depending on the status of the book, they might be re-processed as new books, triggering move operations

Screenshot_20260611_054559
Screenshot_20260611_054613
Screenshot_20260611_054626
Screenshot_20260611_054638
Screenshot_20260611_054721

## Changes
- add API/UI logic/components to merge titles
- refactor logic to process new book after association with title into reusable function
- during merge, re-process books that are not deleted or about to be deleted as those will require manual actions like recover book

This closes #279


backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+                f"Skipping action to process book {book.id} in merge "
+                "operation because book has no current location"
+            )
+            continue
+
+        _book = process_book(session, book, current_location.filename)
+        if _book.location_kind == "prod":
+            prod_book_processed = True
+
+    if prod_book_processed:
```
benoit74: Why should we care about that? apply_retention_rules logic already ensures we consider only prod books, so I would rather call this logic unconditionally and let it decide what needs to be done (or not)

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+    books = [
+        book
+        for book in books
+        if book.location_kind not in ["to_delete", "deleted"]
+        and not book.needs_processing
+        and not book.needs_file_operation
+    ]
+    prod_book_processed = False
+    for book in books:
+        current_location = next(
```
benoit74: Shouldn't we move this "safeguard" logic inside "process_book"? Looking at when this function is called currently, I think we have other edge cases where we should trigger this exit logic.

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+            (
+                loc
+                for loc in book.locations
+                if loc.status == "current" and not loc.is_backup
+            ),
+            None,
+        )
+        if not current_location:
+            logger.warning(
```
benoit74: This is not a warning but a debug log. This is expected and normal to have books which are only in backup, nothing to be warned for.

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+
+    # Restrict action to re-process books to those that have not been deleted, have
+    # non-backup current locations and do not require any action. Books excluded
+    # will require manual action like recover book at which point, the books have
+    # already been attached to the target title
+    books = [
+        book
+        for book in books
+        if book.location_kind not in ["to_delete", "deleted"]
+        and not book.needs_processing
```
benoit74: I don't think that just "ignoring" books which need processing is the right action, it could lead to weird situations. I would rather refuse to merge titles if any source book is in "needs_processing" or "needs_file_operation" status.

WDYT?

backend/src/cms_backend/db/title.py
benoit74: Looking at this twice, I think we should as well move filtering of "deleted" and "to_delete" books in the "process_book" logic.

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
         ),
     )
     return title
+
+
+def merge_titles(session: OrmSession, target: str, srcs: list[str]) -> None:
```
benoit74: For clarity, I would rename:
- `target` => `target_title_name`
- `srcs` => `source_title_names`

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+    """
+    Merge list of titles in src to the target title.
+
+    - Books belonging to the source titles are associated with the target title.
+    - The source titles are deleted along with their collection titles
+    - Books are re-processed as if they are new
+    """
+    try:
+        srcs.remove(target)
```
benoit74: I would rather fail the operation if someone makes a wrong call where target is in sources.

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+    except ValueError:
+        pass
+
+    if not srcs:
+        raise ValueError("No sources to merge from")
+
+    sources = [get_title(session, source) for source in srcs]
```
benoit74: For clarity, `sources` => `source_titles`

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+    dest = get_title(session, target)
```
benoit74: For consistency, we should use either dest or target.

For clarity, I would call this `destination_title` or `target_title`.

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+    books = [book for title in sources for book in title.books]
```
benoit74: source_books

backend/src/cms_backend/db/title.py
```diff
@@ -706,3 +711,68 @@ def revert_title(
...
+    # Attach all the books to the new title
+    for book in books:
```
benoit74: source_book

frontend/src/components/MergeTitlesDialog.vue
```diff
@@ -0,0 +1,141 @@
...
+  >
+    <template #content>
+      <div>
+        <p class="mb-2"><strong>Target Title:</strong> {{ targetTitle }}</p>
+        <p class="mb-2"><strong>Source Title(s):</strong></p>
+        <ul class="ml-4">
+          <li v-for="title in sourceTitles" :key="title">{{ title }}</li>
+        </ul>
+        <v-alert type="warning" variant="tonal" class="mt-4">
+          This action cannot be undone. The source titles will be merged into the target title and
```
benoit74: I would rather say "This action cannot be undone. Books of source titles will be moved to the target title, all books (source and target) will be reprocessed, source titles will be deleted."

frontend/src/components/MergeTitlesDialog.vue
```diff
@@ -0,0 +1,141 @@
...
+    icon-color="primary"
+    :max-width="600"
+    :loading="merging"
+    @confirm="handleMergeConfirm"
+    @cancel="handleMergeCancel"
+  >
+    <template #content>
+      <div>
+        <p class="mb-2"><strong>Target Title:</strong> {{ targetTitle }}</p>
+        <p class="mb-2"><strong>Source Title(s):</strong></p>
```
benoit74: Shall we add a v-if to display source title on a single line when there is only one source (which is the majority of cases tbh)?

frontend/src/components/MergeTitlesDialog.vue
```diff
@@ -0,0 +1,141 @@
...
+  <v-dialog v-model="isTargetDialogOpen" :max-width="600" persistent>
+    <v-card>
+      <v-card-title class="text-h6">
+        <v-icon class="mr-2" color="primary">mdi-merge</v-icon>
+        Select Target Title for Merge
+      </v-card-title>
+
+      <v-card-text class="text-body-1">
+        <p class="mb-4">
+          Select which title should be the target. The other selected titles will be merged into
```
benoit74: I would rather say "Select in which title you want to move all books. Other titles will be deleted."

benoit74 (changes requested):
Lots of small improvements, either I'm rather finicky this morning or you missed many small details 😉

backend/src/cms_backend/db/title.py
elfkuzco: it makes sense.

backend/src/cms_backend/db/title.py
benoit74: Up, looks like you missed the point about book missing any current location

benoit74 (changes requested):

elfkuzco:
When `process_book` is called if book arrives newly, we generate a new name and create target locations based on that name. When we are merging titles, we aren't generating new filenames, just file movement operations. After all, the new books won't conflict with the ones existing in the targe title. Thus, that's why during a merge operation,  I need to get only the first location of the book to get it's current filename. 

But I can add the safeguard to the process_book logic.

EDIT: Or should we generate new names during merge?

benoit74:
> When process_book is called if book arrives newly, we generate a new name and create target locations based on that name. When we are merging titles, we aren't generating new filenames, just file movement operations. After all, the new books won't conflict with the ones existing in the targe title. Thus, that's why during a merge operation, I need to get only the first location of the book to get it's current filename.

OK now I get it, thank you very much for the explanation. 

I think this shows that we should store book filename inside the book. We indeed want to avoid to change book filename in order to ensure stability of user downloads. Not storing this info at book level is probably a mistake. We still need it in the book location because there are edge cases where the filename is different (e.g. when it arrives in quarantine) but it would be easier if we stored at book level its intended (and stable) filename. And this would allows to not have to pass this additional parameter to `process_book`.

WDYT?

elfkuzco:
>   WDYT?

I think it's okay

benoit74 (approved):
LGTM, I like this!

PR #342, opened 2026-06-12
remove id and needs_attention from book filters

### LINKED ISSUE #340: Simplify books screen filters
Now that I'm using a lot the CMS (and that I've seen others using it), I realize we can simplify the book filters

Image

We do not need to filter by ID, only techies would do that and they already know they can input the ID in the URL, I don't see a reasonable scenario where one might want to search for a portion of book ID ; it is also very confusing to not have the "name" as first filter, and it is not consistent with title filters.

The "needs_attention" filter is very annoying because when one want to search for a book which unfortunately needs attention, then it is not returned and one needs to select "Yes" to have the result (blanking the filter does not work). I think it is reasonable as well to remove the filter from this screen, I don't have a reasonable scenario where one might want to filter only for books which do not need attention (we only want to filter for books which do need attention, but this is the inbox screen).

We should remove both "ID" and "needs_attention" filters.

@elfkuzco does this makes sense to you?

elfkuzco:
Screenshot_20260612_102942

## Changes
- remove id and needs_attention parameters from books view
- when needs_attention is not set, use custom ordering for book locations. `quarantine` < `staging` < `prod` < `to_delete` < `deleted`. This way, deleted books are shown last in the books view
- because of custom ordering, remove distinct clause on backup books

This closes #340

### DIFF
--- a/backend/src/cms_backend/db/books.py
+++ b/backend/src/cms_backend/db/books.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from uuid import UUID
 
 from pydantic import AnyUrl
-from sqlalchemy import String, and_, or_, select
+from sqlalchemy import String, and_, case, or_, select
 from sqlalchemy.orm import Session as OrmSession
 
 from cms_backend.context import Context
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM

Issue #322, opened 2026-06-05
Allow to upload manually a ZIM


kelson42:
A publisher (typically not Kiwix) should be able to upload a ZIM via:
* File upload
* URL

Once given, the ZIM file should land in the quarantine and go through the "normal" process.

benoit74:
Some questions:
- are we sure we wanna support file upload? (I don't think API is the right component to support long uploads, so it probably means we need another component)
- don't we wanna restrict publisher to only few titles? how? we probably do not want a publisher to override another publisher titles
- don't we want a more convenient UI for external publishers when it comes to validating their ZIMs? not involving the whole CMS machine which is maybe overkill when publisher is just at the stage of checking his first ZIM matches our constraints ; I'm concerned it might be a burden to external publishers

kelson42:
> Some questions:
> - are we sure we wanna support file upload? (I don't think API is the right component to support long uploads, so it probably means we need another component)

Yes, because once you have (produced) a ZIM manually, this is the straight way to publish it. You might have it available online already... but this is nothing but sure.

I have no problem to limit the upload size to a few GiB in a first version.

> - don't we wanna restrict publisher to only few titles? how? we probably do not want a publisher to override another publisher titles

Nothing here is requested, a publisher should be free within its collection.

> - don't we want a more convenient UI for external publishers when it comes to validating their ZIMs? not involving the whole CMS machine which is maybe overkill when publisher is just at the stage of checking his first ZIM matches our constraints ; I'm concerned it might be a burden to external publishers

Indeed the CMS is not s ZIM developement tool.

This is the role of zimcheck and anyway another topic.

benoit74:
> I have no problem to limit the upload size to a few GiB in a first version.

This is way more than practical for the API. We need another component to support these uploads, well noted.

> Nothing here is requested, a publisher should be free within its collection.

So we have a restriction: each publisher has one collection and must be restricted to it. Are we sure about this? Most readers support only one collection / OPDS catalog ATM, so this requirement would mean we will have new ZIMs ... invisible to mostly everyone. We also do not have good mechanisms in place (yet) to mirror non-Kiwix collections in a straightforward manner

> This is the role of zimcheck and anyway another topic.

Then why are we implementing additional rules in CMS besides zimcheck? I'm quite sure it will be a friction point, but I'm open to consider this is another topic

kelson42:
> This is the role of zimcheck and anyway another topic.
[...]

PR #343, opened 2026-06-12
link title flavours with zimfarm recipe

### LINKED ISSUE #255: Better link Zimfarm recipe and CMS title
Currently, the link between Zimfarm recipe and CMS title is pretty weak: CMS doesn't know which Zimfarm recipe(s) is (are) supposed to create ZIMs for a given title.

Consequences are :
- users can't easily switch from Zimfarm recipe to CMS title or vice-versa
- we can easily end-up where Zimfarm recipe changes the ZIM name (user update of recipe configuration or scraper update), CMS editor assumes a new title needs to be created (instead of updating existing one) and creates a new title with new ZIM name, which ultimately leads to duplicate ZIMs in the catalog
- corrolary of previous point is that CMS cannot help editors when ZIM name changes by suggesting updating the existing title name when it is already associated with same Zimfarm recipe (like say "hey, I have a new ZIM whose name is "xyz" ; the Zimfarm recipe creating this ZIM was previously associated with title "abc" ; do you wanna update "abc" title to "xyz"?)
- CMS cannot detect ZIM name conflicts: two Zimfarm recipes creating ZIMs with same name (and flavour if used) (would solve https://github.com/openzim/zimfarm/issues/1037 when CMS is in use ; pretty sure we have another issue about this topic but I fail to find it)

### LINKED ISSUE #339: add API/UI  to fix recipe issues
Copied from https://github.com/openzim/cms/issues/255#issuecomment-4658245434

When a book has a "recipe issue", we need to allow the user to display a specific screen (I'm now convinced we need one per type of book issue) ; I propose it is a button in book details which open this screen ; we can for instance have the list of issues displayed (not only in a popup) and have a button next to each issue (with only this issue type implemented for the time being).

The screen to solve "recipe issue" must display:

    - current book name + flavour
    - current recipe configuration : title name + list of flavour(s) we expect this recipe to produce

User must be able to either create new title (in which case everything is linked correctly directly) or create/change configuration :

    - select which title this recipe is expected to produce
    - select which flavours this recipe is expected to produce

In both cases (title + flavours) the UI will display a warning if title and/or flavour is already associated with another recipe (with reference to this other recipe info).

UI must present all recipe by their name with a link to Zimfarm UI (by recipe ID to be sure we point to the right object).

When user saves this, a single API call is issued to ensure consistent output. I propose we have something like `PUT /recipes/<recipe_name>` with the title name and the list of flavours as body. This API call will associate given recipe with the title and flavour(s) listed. It will automatically remove other existing associations for this title and flavours with another recipe(s). In order to avoid unattended effects, the list of these other recipes will be given in the request body and must match what API will find in DB. A "title_modified" event is also trigerred to process book again.

### LINKED ISSUE #344: mark books that do not have an existing recipe with `recipe issue`
Copied from https://github.com/openzim/cms/issues/255#issuecomment-4658245434

So looking again at this, I'm think the simplest move is to create two tables:

    a title_flavour table to know for a given title + flavour which recipe is supposed to create it (hence we do not have two recipes creating the same title + flavour)
    a recipe table to know for a given recipe which title it is supposed to create books for (hence we do not have two titles which claim been created by a single recipe)

And let's forget about this producer table and API for the time being, I'm over-engineering the issue on this aspect for now.

Regarding the idea of blocking the notification, I was mostly bought then I realized it means we can end-up in a situation where the ZIM file which does exists on the filesystem is impossible to delete. We should rather block the book.

So when a book is created, CMS will check if recipe exist and is correct:

    if recipe exists and its configuration is matching book name + flavour => proceed
    if recipe exists and its configuration is not matching book name + flavour => block the book in staging without a title associated for the time being, flag as "recipe issue"
    if recipe does not exist => create the recipe (without associated title, but we now know this recipe does exists), block the book in staging without a title associated for the time being, flag as "recipe issue" and flag as "recipe issue"

Note that by recipe configuration I mean both the recipe table and title_flavour table content. In all cases recipe name is updated (looking for recipe by its ID).

elfkuzco:
## Rationale
This PR redefines the way books are attached to titles. When books arrive, their recipe details are retrieved from the zimfarm notification and are only linked to title when the recipe exists in the database and the title (or it's flavoours) to be associated also points to the same same recipe. All other scenarios lead to the books not being associated with a title and an issue `recipe issue` added to it and they are forced to staging. Accordingly, a UI is added to resolve book issues
Screenshot_20260618_073333
Screenshot_20260618_073346
Screenshot_20260618_073403

## Changes
- flag books whose title or it's various flavours do not match recipe from notification with `recipe issues`
- add functions to create zimfarm recipes from notifications
- add logic to update zimfarm recipe with title
- add UI to resolve book issues
- add maintenance script to create zimfarm recipes and title flavours for existing titles
- link recipes with zimfarm API so user can visit 

This closes #344 
This closes #339 
This closes #255


elfkuzco:
@benoit74 , i'm guessing we need a maint-script to populate the db with `ZimfarmRecipe` using ID and names from recipe. Right?

backend/maint-scripts/update_title_flavours_from_books.py
```diff
@@ -37,7 +37,19 @@ def update_title_flavour(session: OrmSession, title: Title) -> tuple[bool, str]:
             session.add(book)
 
     flavours = {book.flavour for book in books if book.flavour is not None}
-    title.flavours = list(flavours)
+    for title_flavour in title.flavours:
```
benoit74: I think it would be easier to simply drop this script. Now that we did run this script to initialize DB when we introduced flavours in titles, we should probably not run it again and I'm really not sure it makes much sense to keep it / maintain it.

backend/src/cms_backend/db/book.py
```diff
@@ -930,3 +933,10 @@ def _recover_deleted_book(session: OrmSession, book: Book) -> Book:
     session.flush()
 
     return book
+
+
+def book_has_recipe_issue(book: Book, recipe: ZimfarmRecipe) -> bool:
+    """Check if book has recipe issues."""
+    if book.name != recipe.name:
```
benoit74: I'm sorry for the confusion, but recipe name has nothing to do with book name. Book name is the ZIM Name. Recipe name is just a convenient label for display in Zimfarm (and CMS now). They often match but this is not mandatory at all.

backend/src/cms_backend/db/models.py
```diff
@@ -269,6 +267,17 @@ class Title(Base):
         order_by="TitleHistory.created_at.desc()",
     )
 
+    zimfarm_recipe: Mapped[Optional["ZimfarmRecipe"]] = relationship(
```
benoit74: This should be a list. A single title can be populated by multiple zimfarm recipes (on per title flavour).

backend/src/cms_backend/db/models.py
```diff
@@ -465,3 +474,40 @@ class Event(Base):
     created_at: Mapped[datetime]
     topic: Mapped[str]
     payload: Mapped[dict[str, Any]]
+
+
+class TitleFlavour(Base):
+    __tablename__ = "title_flavour"
+    id: Mapped[UUID] = mapped_column(
```
benoit74: Why would we need an `id` on this? Can't we use `title_id` + `flavour` as primary key?

backend/src/cms_backend/db/title.py
```diff
@@ -487,6 +503,21 @@ def update_title(
                 f"{getnow()}: locations updated due to title collection change"
             )
 
+    if payload.flavours is not None:
+        # Remove existing flavours
```
benoit74: I think we should not keep updating title flavours through this API. I would rather keep this only for the (yet to be created) `PUT /recipes/<recipe_name>` which ensure we create consistent and complete data.

backend/src/cms_backend/migrations/versions/0c0f13d5733b_create_zimfarm_recipe_and_title_flavour_.py
```diff
@@ -0,0 +1,106 @@
...
+            ["title.id"],
+            name=op.f("fk_title_flavour_title_id_title"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_title_flavour")),
+        sa.UniqueConstraint(
+            "title_id", "flavour", name=op.f("uq_title_flavour_title_id")
+        ),
+    )
+    op.execute(
```
benoit74: I would not do that conversion here, but rather let the maintenance script create records for both zimfarm_recipe and title_flavour tables (we can't do that anyway if recipe_id is not nullable as I suggest it should be).

backend/src/cms_backend/db/models.py
```diff
@@ -465,3 +474,40 @@ class Event(Base):
...
+        init=False, primary_key=True, server_default=text("uuid_generate_v4()")
+    )
+    title_id: Mapped[UUID] = mapped_column(
+        ForeignKey("title.id", ondelete="CASCADE"), init=False
+    )
+    flavour: Mapped[str]
+    recipe_id: Mapped[UUID | None] = mapped_column(
```
benoit74: Let make this not nullable for now. Recent (last friday) discussions with @kelson42 led me to consider we might always have books coming from Zimfarm in the end. And at least this is what we have for now.

backend/src/cms_backend/mill/processors/title.py
```diff
@@ -23,15 +41,50 @@ def add_book_to_title(session: OrmSession, book: Book, title: Title):
...
+                    recipe_name=content["recipe_name"],
+                )
+
+            if book_has_recipe_issue(book, recipe):
+                book.issues = ["recipe issue"]
+            elif recipe.title_id != title.id:
+                book.issues = ["recipe issue"]
+            recipe.name = content["recipe_name"]
+        else:
+            book.issues = ["recipe issue"]
```
benoit74: This is not a recipe issue, this is a very bad event we cannot recover from. We should rather mark this book as errored + add an event explaining the problem

backend/src/cms_backend/mill/processors/title.py
```diff
@@ -23,15 +41,50 @@ def add_book_to_title(session: OrmSession, book: Book, title: Title):
...
-        title.books.append(book)
-        book.events.append(f"{getnow()}: book added to title {title.id}")
-        title.events.append(f"{getnow()}: book {book.id} added to title")
+        if book.zimfarm_notification and not get_missing_keys(
+            book.zimfarm_notification.content, "recipe_id", "recipe_name"
+        ):
+            content = book.zimfarm_notification.content
+            recipe = get_zimfarm_recipe_by_id_or_none(session, content["recipe_id"])
+            if recipe is None:
+                recipe = create_zimfarm_recipe(
```
benoit74: When recipe did not existed yet, we said we need to mark the book as "recipe issue"

backend/src/cms_backend/mill/processors/title.py
```diff
@@ -23,15 +41,50 @@ def add_book_to_title(session: OrmSession, book: Book, title: Title):
...
+                    session,
+                    recipe_id=content["recipe_id"],
+                    recipe_name=content["recipe_name"],
+                )
+
+            if book_has_recipe_issue(book, recipe):
```
benoit74: This whole algorithm has to be revisited to match what I've described around having a matching title + flavour

backend/src/cms_backend/mill/processors/title.py
```diff
@@ -23,15 +41,50 @@ def add_book_to_title(session: OrmSession, book: Book, title: Title):
...
+            recipe.name = content["recipe_name"]
+        else:
+            book.issues = ["recipe issue"]
 
-        # Update title name should it have changed (e.g. stackexchange domain updated
-        # leading to ZIM name automatically updated as well)
-        if title.name != book.name:
-            title.events.append(f"{getnow()}: updating title name to {book.name}")
-            title.name = book.name
+        if not book.issues:
```
benoit74: If book has an issue, they we should log a book event indicating why we did added the book to any title yet.

benoit74 (changes requested):
Yes, we need a maintenance script to populate the DB. This is probably not an easy one to build since it needs to source its data from Zimfarm. Let's try to build something and we will iterate.

backend/src/cms_backend/db/models.py
elfkuzco: I did set an on_delete behaviour for it so that if/when titles are deleted, all the associated title flavours would be deleted.

benoit74: Yes, but why creating a new ID for every title ID + flavour which are already unique? For instance the `book_location` has no ID column because it is not needed.

elfkuzco: was not sure if the ondelete would work with composite primary keys. seems it could work.

backend/src/cms_backend/db/title.py
elfkuzco: I have added the function to update a zimfarm recipe but I realize now it's a bit ambiguous to decide which recipe we want to attach the flavours with during `update_titles` and `revert_title` (from history). I have added a pass statement for now and tests around those would fail while I wait for your opinion on how we should handle it. Should we remove it entirely from this function body?

benoit74: I'm sorry but I don't get what is `update_titles` operation.

Regarding `revert_title`, what if we consider the attachment between recipe and title+flavour(s) is not handled in title history but in recipe history (yet to be added, with corresponding UI btw, can be done in a distinct PR)? This is again more "natural" for the user, easier to display in history UI, and probably way easier to handle. We will still have edge cases like title could be gone, but at least it ensures again consistent updates.

elfkuzco: `update_title` rather, sorry. when user updates title flavours from UI maybe because book says flavour mismatch, which recipe  will the new flavours be attached to? especially since  a single title can be populated by multiple zimfarm recipes (on per title flavour).

benoit74: We should not let users update title flavours from title UI (and hence through `update_title`) anymore indeed, this is ambiguous. Title flavours must be updated through the recipe (assign title + flavours to a recipe) so there is no ambiguities. It means we obviously need to be able to view recipe in UI, edit them, delete them.

elfkuzco:
@benoit74 , I have updated the PR description and added more commits so #339 is resolved through this PR too. I realize they overlap a lot

backend/tests/db/test_zimfarm_recipe.py
```diff
@@ -0,0 +1,164 @@
...
+            dbsession,
+            select(TitleFlavour).where(
+                TitleFlavour.title_id == title.id, TitleFlavour.recipe_id == recipe.id
+            ),
+        )
+        == 3
+    )
+
+
+def test_update_zimfarm_recipe_dissociate_recipes_from_flavours(
```
benoit74: We should also have following test pass:

```
def test_update_zimfarm_recipe_dissociate_recipes_from_flavours(
    dbsession: OrmSession,
    create_zimfarm_recipe: Callable[..., ZimfarmRecipe],
    create_title: Callable[..., Title],
):
    recipe1 = create_zimfarm_recipe()
    title = create_title(zimfarm_recipe=recipe1)
    # "mini", "maxi" and "nopic" flavours belong to recipe1
    mini = TitleFlavour(flavour="mini")
    mini.title = title
    mini.recipe = recipe1

    maxi = TitleFlavour(flavour="maxi")
    maxi.title = title
    maxi.recipe = recipe1

    nopic = TitleFlavour(flavour="nopic")
    nopic.title = title
    nopic.recipe = recipe1

    dbsession.add_all([maxi, mini, nopic])

    # recipe2 just arrives (with mini now produced by it)
    recipe2 = create_zimfarm_recipe()
    dbsession.flush()

    update_zimfarm_recipe(
        dbsession,
        recipe=recipe2,
        flavours=["mini"],
        title=title,
        current_recipes={recipe1.id, recipe2.id},
    )

    # recipe1 still has maxi and nopic
    assert (
        count_from_stmt(
            dbsession,
            select(TitleFlavour).where(TitleFlavour.recipe_id == recipe1.id),
        )
        == 2
    )
    
    # recipe 2 has mini
    assert (
        count_from_stmt(
            dbsession,
            select(TitleFlavour).where(TitleFlavour.recipe_id == recipe2.id),
        )
        == 1
    )

    dbsession.refresh(recipe1)
    assert recipe1.title_id is title.id
    dbsession.refresh(recipe2)
    assert recipe2.title_id == title.id
```

I'm not sure how (see other comment on backend implementation)

backend/src/cms_backend/db/zimfarm_recipe.py
```diff
@@ -0,0 +1,201 @@
...
+def update_zimfarm_recipe(
+    session: OrmSession,
+    *,
+    recipe: ZimfarmRecipe,
+    flavours: list[str],
+    title: Title,
+    current_recipes: set[UUID],
+    create_event: bool = True,
+):
+    """Update a recipe to be associated with the title and flavours.
```
benoit74: This algorithm is not totally accurate and can probably be simplified, see other comment on test case.

Simplification could be to simply iterate over all flavours passed, create or fix corresponding title flavours if needed (recipe is different) and also update recipes (remove title_id from old recipe, add title_id to current recipe) in the same go. And finally, check that all old recipes are not in the list of current_recipes. 

Btw, maybe it would be simpler and more intelligible to simply pass `old_recipes` in the API call, and checking that the old recipes we've collecting during flavours processing do match the passed list.

We do not need to (and should not) touch other flavours not passed in this API call because produced by other recipes, they are "probably" just fine (or another API call will update them)

frontend/src/components/RecipeIssueSection.vue
```diff
@@ -0,0 +1,501 @@
...
+        <v-progress-circular indeterminate size="64" />
+        <div class="mt-4 text-body-1">Loading recipe information...</div>
+      </div>
+
+      <div v-if="!loading && !props.book.recipe_id && !recipeData">
+        <!-- No Recipe Associated -->
+        <v-card variant="elevated" class="mb-4">
+          <v-card-text>
+            <p class="text-body-2 mb-4">
+              Browse and select the Zimfarm recipe that you want to use for this book's
```
benoit74: Why letting user select a recipe? We know this book has been produced by a given recipe, user should not be able to choose a recipe, only fix this book recipe configuration.

Btw, I realize we do not store this information in the book. We probably need to (and this is the reason you had to ask user again for this information).

frontend/src/components/RecipeIssueSection.vue
```diff
@@ -0,0 +1,501 @@
...
+            <p class="text-body-2 mb-4">A new title will be created based on the book metadata</p>
+            <v-btn color="primary" variant="elevated" @click="handleCreateTitle">
+              <v-icon class="mr-2">mdi-plus-circle</v-icon>
+              Create Title
+            </v-btn>
+          </v-card-text>
+        </v-card>
+
+        <v-card v-if="actionMode === 'configure-recipe'" variant="outlined" class="mb-4">
+          <v-card-title>Configure Recipe</v-card-title>
```
benoit74: I would change this to "Modify existing Title". And we should pre-select the title which name matches the book name if it does exists, and pre-select this option (rather than "Create title").

frontend/src/components/RecipeIssueSection.vue
```diff
@@ -0,0 +1,501 @@
...
+                label="Title Name"
+                variant="outlined"
+                density="compact"
+                readonly
+                placeholder="Select a title..."
+              />
+            </div>
+
+            <div class="mb-4">
+              <span class="text-subtitle-1 font-weight-bold">Expected Flavours</span>
```
benoit74: This is really too complex IMHO. And it does not allows to set a blank flavour (book without flavour).

Once title is selected, why not simply:
- listing all current title flavours
- if book flavour is already in title but with another recipe, propose to update title flavour
- and if book flavour is not "known" in current title, propose to add it to the title

And below, that, display consequences on recipes (basically the content of the API request):
- which title the current recipe is associated with
- which flavours the recipe is hence expected to produce (prior ones + current one)
- which recipe will not have anymore the corresponding flavour / title associated

benoit74 (changes requested):
Before reviewing the full PR, I would like to also be able to test it locally.

Did you tried to run the maintenance script on a recent copy of both CMS and Zimfarm? Did it produced meaningful warnings or just too many? We should have only few, or it probably means we have a broader issue.

And how long did it take, is it "reasonable"?

I also feel like the script should take into account known bad flavours like `_maxi` which should be transformed automatically to `maxi` (you can code something like `_<anything>` becomes `<anything>`

I did few tests locally and I feel like "fix recipe" UI should be greatly simplified, see inline comments.

Besides this PR, we will need more screen to manage existing recipes (outside of any book issue).

elfkuzco:
> Did you tried to run the maintenance script on a recent copy of both CMS and Zimfarm? Did it produced meaningful warnings or just too many? We should have only few, or it probably means we have a broader issue.

I ran the script on prod zimfarm but i didn't let it complete because there were too many. If I ran it locally, i would have had more than 10k tasks across 2k recipes. Most of it are just warnings because title flavour already exist. I couldn't run locally because for some reason, i can't have two database containers in two different folders running . Don't know if it has to do with the different postgres versions of CMS and Zimfarm

backend/tests/db/test_zimfarm_recipe.py
elfkuzco: Thank you very much for the test. It expose a flaw in my db model and the way I link title with recipe.

benoit74:
> I ran the script on prod zimfarm but i didn't let it complete because there were too many. If I ran it locally, i would have had more than 10k tasks across 2k recipes. Most of it are just warnings because title flavour already exist.

These are not warnings but debug logs. I was specifically referring to warnings like abnormal unexpected things.

> I couldn't run locally because for some reason, i can't have two database containers in two different folders running . Don't know if it has to do with the different postgres versions of CMS and Zimfarm

I can run both fine on my machine, side-by-side.

We need to really find a solution to run this migration (and any migration in fact) before releasing to production. We can't just hope the migration will have the intended effect, especially on very large migrations like this one which could have multiple bugs linked to forgotten conditions between Zimfarm and CMS status.

elfkuzco:
> I can run both fine on my machine, side-by-side.

Finally fixed it with the `-p` flag to `docker compose`. It appears that without it, my database containers crash

https://github.com/docker-library/postgres/issues/1363

backend/tests/db/test_zimfarm_recipe.py
elfkuzco: Fixed now

backend/src/cms_backend/db/zimfarm_recipe.py
elfkuzco: Fixed now

frontend/src/components/RecipeIssueSection.vue
elfkuzco: Screenshot_20260618_232257

WDYT?

elfkuzco:
> We need to really find a solution to run this migration (and any migration in fact) before releasing to production. We can't just hope the migration will have the intended effect, especially on very large migrations like this one which could have multiple bugs linked to forgotten conditions between Zimfarm and CMS status.

Ran the script and so far, the only warnings I have been getting on my copy of zimfarm DB is:

```sh
[2026-06-19 00:04:41,589: WARNING] Task a60ba59e-8f17-47ed-8899-2064e1f4a8c0 metadata is missing keys: Name
[2026-06-19 00:04:41,606: WARNING] Task 2ff2e7e8-204e-4b61-8f5c-6e16da10e0e7 metadata is missing keys: Name
[2026-06-19 00:04:41,623: WARNING] Task 667ba7ab-98d9-4a89-88f0-7a6b0052f4f0 metadata is missing keys: Name
[2026-06-19 00:04:41,640: WARNING] Task 75eb9de0-41c9-462b-826a-6ac60202eb3f metadata is missing keys: Name
```

elfkuzco:
> We need to really find a solution to run this migration (and any migration in fact) before releasing to production. We can't just hope the migration will have the intended effect, especially on very large migrations like this one which could have multiple bugs linked to forgotten conditions between Zimfarm and CMS status.

Ran the script and so far, the only warnings I have been getting on my copy of zimfarm DB is:

```sh
[2026-06-19 00:04:41,589: WARNING] Task a60ba59e-8f17-47ed-8899-2064e1f4a8c0 metadata is missing keys: Name
[2026-06-19 00:04:41,606: WARNING] Task 2ff2e7e8-204e-4b61-8f5c-6e16da10e0e7 metadata is missing keys: Name
[2026-06-19 00:04:41,623: WARNING] Task 667ba7ab-98d9-4a89-88f0-7a6b0052f4f0 metadata is missing keys: Name
[2026-06-19 00:04:41,640: WARNING] Task 75eb9de0-41c9-462b-826a-6ac60202eb3f metadata is missing keys: Name
```


Took 3hrs in total

benoit74:
Thank you, great news for the migration script!

Regarding API / UI, I realize again this is just really complex. I'm sorry but I will need some time to think twice again about this change.

elfkuzco:
Screenshot_20260619_164515


updated maint-script and api/ui to view recipe history

benoit74:
We've discussed the matter with @kelson42 last Friday and the output is that he considers our approach is significantly wrong.

His main point was that we should not ask users to re-enter things which can be assumed from book content ; we should just move the book to quarantine or staging, and have buttons to automatically do what is needed. For instance it means that for association of book with title, we should have a "set title" button on the book which proposes to either choose an existing title or create a new one, and automatically assume everything coming from the book is correct once user has chosen the correct title.

I expect all this should be done in the backend, preferably in a single API call for consistency, but then I'm not sure what to show on the UI so that user still knows what he asks the backend to do.

A second point what that we should not show / speak about Zimfarm recipe in the CMS since it is a Zimfarm attribute. I absolutely don't know what to do from this information.

A third point was this is plain simple to him and he does not understand why we are speaking about the matter. I don't know what to do of this statement.

@elfkuzco do you know how to move forward from these statements or do we need to arrange yet another discussion with @kelson42

elfkuzco:
> For instance it means that for association of book with title, we should have a "set title" button on the book which proposes to either choose an existing title or create a new one, and automatically assume everything coming from the book is correct once user has chosen the correct title.

I believe we already have a button to add book to title and buttons to create title. Just not shown in the issue tab.

> A second point what that we should not show / speak about Zimfarm recipe in the CMS since it is a Zimfarm attribute. I absolutely don't know what to do from this information.

Given the issue was to link zimfarm recipe and cms title, if we don't display the information in the API/UI, i'm not sure how issues would be fixed. For example, if this information is completely hidden and we create recipes as they arrive, we can have a conflict where a book arrives and we don't create a recipe because the recipe already exists. But the recipe points to a different title from the one than the matching title. Such books would just remain with `reicpe issues` without getting fixed (assuming the recipe details are to be completely hidden)

benoit74:
After discussing with @kelson42  and after some thought, I consider that to solve this issue correctly, we need to:
- First complete https://github.com/openzim/cms/issues/395, this is what we need to "set title" (which was incorrect words)
- Then consider title flavour / recipes issues just as another form of validation action (user just has to validate that he is OK with the given recipe producing given flavour for given title)
- Drop `zimfarm_recipe` DB table (and associated objects) 
  - This table is mainly needed to ensure we do not have two titles being produce by a single recipe
  - This is not a CMS problem per se (more a Zimfarm configuration issue)
  - We could decide to evolve Zimfarm to allow that
  - We can anyway have reporting on that kind of oddities or introduce that later
  - As a consequence, we will not display recipes in the UI and not have an API for them (sorry for that back-and-forth) 
  - And we will query Zimfarm when we need to display recipe name or other stuff like that in the CMS UI (and fallback to recipe ID if Zimfarm is unavailable which is still OK)

All changes regarding recipes is because after discussing with @kelson42 he is too worried we end-up with configuration about recipes in CMS ... which should not happen except association with title ; and I agree about the danger.

The validation will consist in considering that book is "correct" with multiple actions induced (add flavour, associate recipe, etc ...)

elfkuzco:
> Then consider title flavour / recipes issues just as another form of validation action (user just has to validate that he is OK with the given recipe producing given flavour for given title)

> Drop zimfarm_recipe DB table (and associated objects) 

> The validation will consist in considering that book is "correct" with multiple actions induced (add flavour, associate recipe, etc ...)

@benoit74 , please, could you provide guidance on how to achieve this? I am bit confused because if we are going to drop `zimfarm_recipe` DB table, doesn't that mean we don't want to store any information about recipe on CMS? If that's the case, we can't detect a recipe issue, right?

benoit74:
We will continue to store `recipe_id` in title_flavour and in book tables. But only this ID.

elfkuzco:
@benoit74 , I have force-pushed the branch from a completely different state and tree is cleanly based on main. Here's a summary of the updated changes: 
- add maint-script to create new title flavour objects from ZIMFARM_API. You will have to set the zimfarm API `INFORM_CMS` to `false` so that it doesn't try to contact cms for the view/download URLs. It greatly reduces the time to completion
- when a new book is to be processed, check if it has a zimfarm notification as not all books will have a notification (https://github.com/openzim/cms/issues/322). Additionally, create a title flavour object from this book's information if none exists. This means, we will never have `flavour mismatch` errors from new book arrival but this can happen on book update till https://github.com/openzim/cms/issues/396 is completed.
- detect if a book has `recipe issue` if it's notification's reicpe id does not exist in the current flavours
- remove `flavours` details from title create/update payload
- optionally filter book flavours by title_id. This will be used in the book edit tab to ensure we are only going to change the flavour to a supported flavour and not the complete list of flavours
- link recipe detail in book view 
Screenshot_20260709_135116

backend/src/cms_backend/db/book.py
```diff
@@ -729,10 +752,21 @@ def get_book_issues(
...
         issues["flavour mismatch"] = [
             f"book flavour {book.flavour} is not in list of "
-            f"title flavours: {','.join(book.title.flavours)}"  # pyright: ignore[reportOptionalMemberAccess]
+            f"title flavours: {','.join(title_flavours)}"
+        ]
+
+    if book.zimfarm_notification and book_has_recipe_issue(book, book.title):  # pyright: ignore[reportArgumentType]
+        title_recipes = [str(tf.recipe_id) for tf in book.title.flavours]  # pyright: ignore[reportOptionalMemberAccess]
+        recipe = book.zimfarm_notification.content.get(
+            "recipe_name", book.zimfarm_notification.content.get("recipe_id")
```
benoit74: Drop recipe_name here as well

backend/src/cms_backend/db/book.py
```diff
@@ -106,6 +107,23 @@ def create_book_full_schema(book: Book) -> BookFullSchema:
         if location.status == "target"
     ]
 
+    if book.title:
```
benoit74: Please rather store recipe_id in book table, populate this column from zimfarm notification when available, and use it when querying book data like here.

backend/src/cms_backend/db/book.py
```diff
@@ -129,7 +147,9 @@ def create_book_full_schema(book: Book) -> BookFullSchema:
         current_locations=current_locations,
         target_locations=target_locations,
         title_archived=book.title.archived if book.title else False,
-        has_flavour_mismatch=has_flavour_mismatch(book.flavour, book.title.flavours)
+        has_flavour_mismatch=has_flavour_mismatch(
```
benoit74: Ultimately we should not need this property anymore, this is just another kind of issue to solve with "promote_book" action. If this is needed for UI, then please open an issue so we discuss this there

backend/src/cms_backend/db/book.py
```diff
@@ -284,15 +305,28 @@ def move_book(
         raise RecordDoesNotExistError(
             f"Book {book_id} does not meet criteria to be moved."
         )
+
```
benoit74, line 288: Why moving these checks outside of `move_book_to_destination`? The promote action is the only other caller of `move_book_to_destination`, and when promoting I expect we should still check for these (they should have been fixed before promotion). What do I miss?

backend/src/cms_backend/db/book.py
```diff
@@ -729,10 +752,21 @@ def get_book_issues(
             f"{','.join(different_metadata_keys)}"
         ]
 
-    if has_flavour_mismatch(book.flavour, book.title.flavours):  # pyright: ignore[reportOptionalMemberAccess]
+    title_flavours = get_title_flavours(book.title)  # pyright: ignore[reportArgumentType]
```
benoit74: I don't feel like we need all these pyright hints. Are they really needed?

backend/src/cms_backend/db/book.py
```diff
@@ -729,10 +752,21 @@ def get_book_issues(
...
+    if has_flavour_mismatch(book.flavour, title_flavours):  # pyright: ignore[reportOptionalMemberAccess]
         issues["flavour mismatch"] = [
             f"book flavour {book.flavour} is not in list of "
-            f"title flavours: {','.join(book.title.flavours)}"  # pyright: ignore[reportOptionalMemberAccess]
+            f"title flavours: {','.join(title_flavours)}"
+        ]
+
+    if book.zimfarm_notification and book_has_recipe_issue(book, book.title):  # pyright: ignore[reportArgumentType]
```
benoit74: We should be more consistent between `has_flavour_mismatch` and `book_has_recipe_issue`:
- naming is not consistent
- file location is not consistent
- in `has_flavour_mismatch` we pre-compute all arguments so that we know how to display them in the issue, while in `book_has_recipe_issue` we compute some (the list of title recipes) in 

Is there really a reason for all these differences? On a personal basis, I tend to prefer to have everything in `db/book.py`, pass only the book as argument and return issue message if any issue (pre-computed inside the function so we can easily test for it) or None if there is no issue. I'm open to other code design, but this needs to be consistent for maintainability.

backend/src/cms_backend/db/book.py
```diff
@@ -1160,3 +1194,12 @@ def _recover_deleted_book(session: OrmSession, book: Book) -> Book:
     session.flush()
 
     return book
+
+
+def book_has_recipe_issue(book: Book, book_title: Title) -> bool:
+    """Check if book has recipe issues."""
+    if not book.zimfarm_notification:
+        return False
+    return book.zimfarm_notification.content.get("recipe_id") not in [
```
benoit74: We don't want to check if any title flavour has proper recipe. 

We want to check if title flavour matching book flavour has proper recipe.

backend/src/cms_backend/db/book_actions.py
```diff
@@ -164,18 +161,6 @@ def get_book_promotion_actions(
                 message="Configure title collection(s)",
             )
         )
-    if has_flavour_mismatch(book.flavour, title.flavours):
```
benoit74, line 167: Why dropping this? We can still have a book with flavour mismatch. And we miss a new action for recipe issues (title already has flavour but recipe is not matching)

backend/src/cms_backend/mill/processors/title.py
```diff
@@ -36,6 +40,24 @@ def add_book_to_title(session: OrmSession, book: Book, title: Title):
...
+        if book.zimfarm_notification:
+            missing_recipe_keys = get_missing_keys(
+                book.zimfarm_notification.content, "recipe_id", "recipe_name"
+            )
+            if missing_recipe_keys:
+                raise ValueError(
+                    "book notification is missing recipe details: "
+                    f"{','.join(missing_recipe_keys)}"
+                )
+            title_flavour = get_title_flavour_or_none(session, title.id, book.flavour)
```
benoit74: Why adding a title flavour only if book has a zimfarm notification? We need flavours also for books without zimfarm notification.

backend/src/cms_backend/db/models.py
```diff
@@ -282,6 +279,23 @@ class Title(Base):
...
+    )
+
+
+class TitleFlavour(Base):
+    __tablename__ = "title_flavour"
+    title_id: Mapped[UUID] = mapped_column(
+        ForeignKey("title.id", ondelete="CASCADE"), init=False, primary_key=True
+    )
+    flavour: Mapped[str] = mapped_column(primary_key=True)
+    recipe_id: Mapped[UUID]
```
benoit74: Since zimfarm notification is optional, then recipe_id should be optional as well

backend/src/cms_backend/db/title.py
```diff
@@ -349,7 +350,7 @@ def create_title_history_entry(
         source=title.source,
         maturity=title.maturity,
         archived=title.archived,
-        flavours=title.flavours,
+        flavours=get_title_flavours(title),
```
benoit74: We need to also have the recipe ID stored in history. No need to recover previous history entries, but new history entries should adopt new format with flavour + recipe_id (if present) recorded

backend/src/cms_backend/db/title.py
```diff
@@ -396,7 +397,9 @@ def update_title(
         raise RecordDoesNotExistError("Title is not archived.")
 
     update_data = payload.model_dump(
-        exclude_unset=True, exclude={"collection_titles", "comment"}, mode="json"
+        exclude_unset=True,
+        exclude={"collection_titles", "comment", "flavours"},
```
benoit74, line 402: I agree we should not update "flavours" from here anymore, this will be solved in https://github.com/openzim/cms/issues/396, but payload does not seem to include flavours anymore.

And how do we manage flavours until https://github.com/openzim/cms/issues/396 is solved? Are we in a position where we can't update flavours anymore? (I'm OK with that if we agree this is probably the next issue to solve)

backend/src/cms_backend/schemas/orms.py
```diff
@@ -65,6 +73,7 @@ class TitleHistorySchema(TitleLightSchema):
     author: str
     collections: list[BaseTitleCollectionSchema]
     created_at: datetime
+    flavours: list[str]
```
benoit74: As mentioned above, should be `list[TitleFlavourSchema]` instead of `list[str]`

backend/maint-scripts/update_title_flavours_from_zimfarm.py
```diff
@@ -0,0 +1,230 @@
...
+            session.flush()
+            logger.info(f"Created title flavour '{flavour}' for title '{title.name}'")
+
+        # update books notifications whose flavour matches
+        books = session.scalars(
+            select(Book).where(Book.flavour == flavour, Book.title_id == title.id)
+        ).all()
+        for book in books:
+            if book.zimfarm_notification and get_missing_keys(
+                book.zimfarm_notification.content, "recipe_id", "recipe_name"
```
benoit74: Let's consider recipe_name is not mandatory anymore, we do not store it anyway

backend/src/cms_backend/mill/processors/title.py
```diff
@@ -36,6 +40,24 @@ def add_book_to_title(session: OrmSession, book: Book, title: Title):
         if not book.date:
             raise Exception("book date is missing or invalid")
 
+        if book.zimfarm_notification:
+            missing_recipe_keys = get_missing_keys(
+                book.zimfarm_notification.content, "recipe_id", "recipe_name"
```
benoit74: Drop recipe_name, we do not store it anyway

benoit74 (changes requested):
Thank you.

Main feedback is that we should not populate DB at deployment by creating title flavours only from Zimfarm information, since we already have that information in flavours column of title table. It misses the recipe ID which we indeed need to fetch from Zimfarm. The migration script should do the migration from title column to title_flavour row, and the current maint-script is mostly ok to populate recipe_id afterwards, except that it should not create missing title flavours (too risky with very old tasks which are not accurate anymore) and it should update recipe_id if current flavour already exists **and** recipe_id is not set. If recipe_id is already set for given flavour, it should display a warning (should we have two recipes targeting same title + flavour).

backend/src/cms_backend/db/book.py
elfkuzco: `move_book` does prevents move operation when there is a flavour mismatch whereas `promote_book` doesn't (after all, it can't solve flavour issues anymore). So, I have moved the checks to the `move_book` and kept the `move_book_to_destination` simpler

backend/src/cms_backend/db/book.py
elfkuzco: Yeah, otherwise, I am going to be making checks and raise ValueError if book has no title. Places where it's used already make the assumption that book has a title

backend/src/cms_backend/db/book.py
benoit74: Why `promote_book` would not solve flavour issues anymore? It should IMHO, what do I miss?

backend/src/cms_backend/mill/processors/title.py
elfkuzco: But we aren't going to be creating TitleFlavours with recipe_id set to NULL, are we?

elfkuzco: Never mind, seen latest comment on it being optional

backend/src/cms_backend/db/title.py
elfkuzco: We could but I'm not sure if it's ideal to do so. For instance, if we allow user to create/update a title flavour, will the payload for flavours require `recipe_id`, so, it won't be a list of strings but a list of dictionary? If it becomes a list of dictionary, will that really be convenient for user as they would need to copy recipe_id?

If we stick with list of strings as it is, what happens when user creates a new TitleFlavour in the hopes of solving a `flavour mismatch` but a later notification comes from zimfarm with the same flavour but with a recipe_id?

backend/src/cms_backend/db/book_actions.py
elfkuzco: With new logic to create `TitleFlavour` when notification comes in, it means new books will never have `flavour mismatch` unless this will be introduced by an editor perhaps when they edit something. Part of my rationale for removing it is expressed in https://github.com/openzim/cms/pull/343#discussion_r3555497174 .

Given this, if we still want to solve flavour mismatch issues, it means we will have to wait for API for #396 because tilte update doesn't accept flavours as list of strings

backend/src/cms_backend/db/book.py
elfkuzco: See https://github.com/openzim/cms/pull/343#discussion_r3555527928

backend/src/cms_backend/db/title.py
benoit74: When creating / updating flavours, users should indeed only pass a list of flavours (i.e. list of strings) without any information about recipe ID, this is too complex.

When a flavour already exists in DB but its `recipe_id` is empty, then we need to ensure there is no flavour mismatch (flavour is OK) but a recipe mismatch (we need CMS publisher to confirm this is the intended recipe which produced the given flavour).

backend/src/cms_backend/db/book_actions.py
benoit74: We should not automatically create title flavours when notification comes in, this is probably something I've overlooked, sorry. We still want to solve flavour mismatch issues. I'm OK to fait for #396 to fully solve the flavours handling (should be next issue to solve on the list once this is merged).

benoit74:
I think most of the confusion around "flavour mismatch" comes from the fact that I've been a bit imprecise. 

We should not let user create flavours manually anymore in the UI by typing / selecting flavour name as mentioned in https://github.com/openzim/cms/issues/396, because it involves too much risk of typos / mistakes. 

But we still need to let user confirm flavour + recipe when a new one arrives from Zimfarm, so that we do not have "unattended" flavours or recipes automatically configured. The distinction is subtle but important.

So to rephrase:
- we do not have a UI to create new title flavour
- but we still have flavour mismatch + recipe mismatch when a ZIM with a new flavour + recipe arrives
- and this is solvable only through book promotion action

Is this clearer?

elfkuzco:
Yes, I understand most of it. But if we don't create title flavours when notification comes in as mentioned here: https://github.com/openzim/cms/pull/343#discussion_r3585286374 and we don't let user create title flavours, how will we solve flavour mismatch issues? I was thinking logic to promote_book would find an existing title flavour object and use it to update book's flavour. Obviously, we can't create one anymore since we don't have `recipe_id`.

benoit74:
>  I was thinking logic to promote_book would find an existing title flavour object and use it to update book's flavour. Obviously, we can't create one anymore since we don't have recipe_id.

I'm sorry I don't get why we can't create one anymore during book promotion (inside the `/promote` endpoint handling). We will have a recipe_id at that time since book will have that info. And even if recipe_id is missing (because book was not created from a zimfarm notification), this should not be a problem preventing to create a title flavour.

elfkuzco:
Okay

elfkuzco:
Screenshot_20260715_123402
Screenshot_20260715_123105


Refactored book_actions to create title flavour to solve book mismatch

benoit74:
Please change our convention from `dev_` to `dev-` or `dev.` for all objects created with dev scripts so that we have a single convention, this is going to be simpler to maintain.

benoit74:
Edit: I've updated my previous comment, I reviewed last changes too fast, my bad

benoit74 (changes requested):
Some more little changes:
- it looks like we miss a book promotion action to only update flavour recipe when the flavour already exists but recipes do not match (i.e. book has a "recipe issue") ; currently despite book issue, book promotion says there is nothing wrong to move book to production (which is wrong ^^)
- the "setup_notification.py" script assumes "zimcheck.json" is in current working directory while it should not ; it should fetch current script location (with `Path(__file__).parent`) and assume "zimcheck.json" is in same folder
- the "backend/maint-scripts/update_title_flavours_from_zimfarm.py" script should commit changes after each recipe is processed, so that we do not put a lock on all DB titles as script is progressing + should something bad happen we do not loose all changes already computed ; I feel like doing it at recipe level is OK

I've identified other small enhancements we probably wanna make but I will keep them for distinct issues so that we have a better chance to merge this PR soon.

I've doubled check the issue I spoke about yesterday regarding DB population with empty flavours and in fact everything is fine besides my DB viewer which doesn't display correctly arrays with a single empty value ^^ Sorry for the noise

elfkuzco:
> it looks like we miss a book promotion action to only update flavour recipe when the flavour already exists but recipes do not match (i.e. book has a "recipe issue") ; currently despite book issue, book promotion says there is nothing wrong to move book to production (which is wrong ^^)

Rationale for skipping it is if we are fixing it, we have to use either book recipe_id or existing title flavour recipe_id. If we use the existing title, won't it mask the fact that zimfarm has an issue because it has two recipes producing one flavour for the title? If we use the book recipe_id to update, then, other books existing would have `recipe_issue` because their recipe_id is different. Which approach should I use?

elfkuzco:
Screenshot_20260716_122447

benoit74:
> Rationale for skipping it is if we are fixing it, we have to use either book recipe_id or existing title flavour recipe_id. If we use the existing title, won't it mask the fact that zimfarm has an issue because it has two recipes producing one flavour for the title? If we use the book recipe_id to update, then, other books existing would have recipe_issue because their recipe_id is different. Which approach should I use?

The fact that two recipes might be producing the same book will indeed not be solved by this PR, this is intentional and maybe outside CMS duty. 

And just like any other change on title, the fact that one book update some title properties might indeed induce other books to have new issues, this is a "known issue".

benoit74 (approved):
LGTM, thanks a lot for this very difficult PR.

PR #345, opened 2026-06-15
accurately highlight correct navbar item

### LINKED ISSUE #341: Navigation bar selected item is confusing
When you click "Titles", the navigation bar selects the "Titles" entry which is correct.

The problem is that this is never updated until you click on another navigation bar item. 

For instance, if you select a title, "Titles" is still selected which is correct.

But then if you open a book of this title then "Titles" is still selected which is very confusing. 

Image

Having "Books" selected would greatly help the user know he is currently viewing a book.

I assume we could have sensible navigation item to select for every Vue route we have defined. Or maybe the UI should have something else more clearly indicating which kind of object we are seeing, but I feel like the navigation bar update is the most promising solution at this stage.

elfkuzco:
Screenshot_20260615_105858
Screenshot_20260615_105841
Screenshot_20260615_105820
Screenshot_20260615_105805

## Changes
- add `parentNavigation` metadata to routes
- detect the correct navigation entry to highlight based on the parent navigation or the route name

This fixes #341

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -56,7 +56,7 @@ const routes = [
     path: '/books/backups',
     name: 'backup-books',
     component: BackupBooksView,
-    meta: { title: 'CMS | Backup Books' },
+    meta: { title: 'CMS | Backup Books', parentNavigation: 'books' },
   },
   {
     path: '/titles',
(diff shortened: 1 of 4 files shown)


frontend/src/router/index.ts
```diff
@@ -130,6 +135,7 @@ const routes = [
     component: UserView,
     props: true,
     meta: {
+      parentNavigation: 'users-list',
```
benoit74: Probably worth to rename this route to `users` instead of `users-list` for consistency with other lists routes naming

benoit74 (approved):
LGTM, I let you decide if you wanna fix my comment and merge on your own

frontend/src/router/index.ts
elfkuzco: Done

PR #352, opened 2026-06-22
keep prod books in prod despite reprocessing

### LINKED ISSUE #349: Merging titles should not cause book quality to be reassessed
We've just merged two titles with @Popolechien and it looks like it caused book quality to be reassessed, and forced a book to go back from production to staging.

I doubt we really want to do this, since it cause sudden move of a production book to staging, which broke ongoing downloads (if any).

@elfkuzco do you recall if this was intentional or just accidental?

elfkuzco:
## Changes
- keep books in prod if they are already in prod despite issues being re-computed 

This fixes  #349


benoit74 (approved):
LGTM

PR #354, opened 2026-06-22
check book Name metadata meets constraints

### LINKED ISSUE #9: ZIM file format should be checked
Should match something like `/^(.+?_)([a-z\-]{2,10}?_|)(.+_|)([\d]{4}|)\-([\d]{2})$/i`

elfkuzco:
## Rationale
This PR adds a check for the book's Title metadata and passes the title name to the logic to compute book filename insead of the book name.

## Changes
- define regex for valid title names
- mark books whose `Name` metadata fails to match regex with `bad metadata`
- extend UI rules in TitleForm to validate input with regex
- use title name in computing book filename instead of book name

This closes #9

### DIFF
--- a/backend/src/cms_backend/utils/zim.py
+++ b/backend/src/cms_backend/utils/zim.py
@@ -1,4 +1,9 @@
+import re
 from typing import Any
+
+ZIM_TITLE_NAME_REGEX = re.compile(
+    r"^[a-z0-9\-\.]+?_[a-z]{2}(?:-[a-z]{2,10})?_[a-z0-9\-\.]+?$"
+)
 
 
 def get_missing_keys(payload: dict[str, Any], *keys: str) -> list[str]:
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 from typing import Any, Literal
 from uuid import UUID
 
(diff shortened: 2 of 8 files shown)


elfkuzco:
@benoit74 , I stripped down the regex defined by @kelson42  in https://github.com/openzim/cms/issues/9#issue-896332358 as the logic to compute book filename adds flavour and the date parts to the name.

backend/src/cms_backend/utils/zim.py
```diff
@@ -1,5 +1,8 @@
+import re
 from typing import Any
 
+ZIM_TITLE_NAME_REGEX = re.compile(r"^(.+?_)([a-zA-Z\-]{2,10}?_|)(.+)$")
```
benoit74: I just realize this is not the correct regex, this is too permissive. We've agreed in https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention that parts must contain way less characters.

`^[a-z0-9\-\.]+?_[a-z]{2}(?:-[a-z]{2,10})?_[a-z0-9\-\.]+?$` seems more accurate

benoit74 (approved):
Please fix regex and you can merge.

backend/src/cms_backend/utils/zim.py
elfkuzco: Done

PR #355, opened 2026-06-23
add task to delete zimcheck results from s3

### LINKED ISSUE #113: Cleanup zimcheck S3 bucket
This is a followup to https://github.com/openzim/zimfarm/issues/1456

We need a background job to delete zimcheck files from S3 bucket when they are not needed anymore.

Since CMS is primary "user" of zimcheck data, it seems to make more sense the CMS is responsible to delete useless zimcheck files.

Once CMS has decided that zimcheck data is not needed anymore, probably the Zimfarm does not need it anymore either. 

Zimfarm probably doesn't need to have read access to this data (it is only computing and writing this data, accessing it is more a convenience).

elfkuzco:
## Changes
- add background shuttle task to delete zimcheck results from s3. results are deleted only if the zimcheck summaries have been computed and book is not in quarantine

This closes #113


backend/src/cms_backend/shuttle/context.py
```diff
@@ -44,3 +44,6 @@ class Context:
     )
 
     local_warehouse_paths: ClassVar[dict[UUID, Path]] = _parse_local_warehouse_paths()
+    zimcheck_results_upload_uri: str = os.getenv(
```
benoit74: zimcheck_results_upload_uri => zimcheck_results_s3_bucket_uri

backend/src/cms_backend/shuttle/delete_zimcheck_s3_results.py
```diff
@@ -0,0 +1,138 @@
...
+
+    skip = 0
+    limit = 50
+    while True:
+        books = session.scalars(
+            select(Book)
+            .where(
+                Book.zimcheck_result_url.is_not(None),
+                Book.zimcheck_s3_deleted.is_(False),
+                Book.location_kind.not_in(["quarantine"]),
```
benoit74: We want to remove file from S3 only if book is in production or deleted. We wanna keep it (for further manual analysis if needed) when book is in quarantine or staging.

backend/src/cms_backend/db/models.py
```diff
@@ -157,6 +157,10 @@ class Book(Base):
         default_factory=dict, server_default="{}"
     )
 
+    zimcheck_s3_deleted: Mapped[bool] = mapped_column(
```
benoit74, line 160: We probably need a partial index on `zimcheck_s3_deleted = False` to efficiently retrieve these few books which still have zimcheck results on S3 (there will be very few once this PR is merged)

backend/src/cms_backend/shuttle/delete_zimcheck_s3_results.py
```diff
@@ -0,0 +1,138 @@
...
+        upload_uri = urllib.parse.urlparse(ShuttleContext.zimcheck_results_upload_uri)
+        pathlib.Path(upload_uri.path)
+    except Exception as exc:
+        raise ValueError(
+            f"invalid upload URI: `{ShuttleContext.zimcheck_results_upload_uri}`"
+        ) from exc
+
+    s3 = get_kiwix_storage_client(upload_uri)
+
+    logger.info("Deleting zimcheck resultss from S3")
```
benoit74: resultss => results

backend/src/cms_backend/shuttle/delete_zimcheck_s3_results.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if not books:
+            logger.info(
+                "No more books meet criteria for zimcheck results to be deleted."
+            )
+            break
+
+        for book in books:
+            # Determine whether we need to fetch zimcheck results or not
+            missing_summary_keys = get_missing_keys(
```
benoit74: Why should we care about that? I'm concerned we will end with many zimcheck results not been deleted only because we failed to compute zimcheck summary. Once in production, book should have a summary and anyway we don't care much about quality anymore.

benoit74 (changes requested):

benoit74 (approved):
LGTM, I've created necessary user on Wasabi with (hopefully) correct permissions + created the env variable

PR #356, opened 2026-06-23
move limits and skip for task outside while loop

elfkuzco:
## Changes
- move limit and skip variables outside while loop in task to mark staging books for deletion


benoit74 (approved):
Nice catch!

PR #359, opened 2026-06-24
show message when book has pending operations

### LINKED ISSUE #357: Indicate that book buttons are hidden because operation is pending
When a book has a pending operation, buttons are hidden in book details view. This is correct but could be confusing to the (inexperienced) user (I'm always confused too).

We should simply show something saying operations are pending and we should wait to start new things.

elfkuzco:
Screenshot_20260624_092524

## Changes
- add new computed property for detecting if book has pending operations
- show alert message when action buttons are hidden because book has pending operations

This closes #357

### DIFF
--- a/frontend/src/views/BookView.vue
+++ b/frontend/src/views/BookView.vue
@@ -75,6 +75,22 @@
       </div>
 
       <v-alert
+        v-if="hasPendingOperations"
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #360, opened 2026-06-24
show book issues in UI

### LINKED ISSUE #358: Better UI to view book issues
Currently, book issues are rather hard to diagnose.

For each kind of book issue, we should have a dedicated UI (all in same page, like in https://github.com/openzim/cms/pull/343, and collapsible) showing details of the reason the issue arised. 

In order to not transfer some "intelligence" to the UI (e.g. for "media count" issue, UI should not automatically fetch previous book to display its media count) but rather add another API to retrieve details about all book issues (which could also allow to check issue is still valid), which are in turn displayed in the UI.

In most cases, user will not be able to act on the issue besides moving book to production ; this is intended.

elfkuzco:
## Rationale
This PR enhances the API to be able to view the list of book issues in the book detail view
Screenshot_20260624_155912
Screenshot_20260624_154542
Screenshot_20260624_154447
Screenshot_20260625_130150

## Changes
- refactor the logic to update book issues to use the results of new function `get_book_issues` which computes the issues
- add issues tab in UI
- add API to retrieve book issues

This closes #358

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -561,102 +561,71 @@ def revert_book(
     return book
 
 
-def book_has_bad_metadata(book: Book, *, update_events: bool = False) -> bool:
-    """Determine if a book has bad metadata.
-
-    Assumes book has all mandatory metadata
-    """
+def get_book_metadata_issues(book: Book) -> list[str]:
+    issues: list[str] = []
     name = book.zim_metadata["Name"]
+
     if not re.match(ZIM_TITLE_NAME_REGEX, name):
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Name metadata contains unwanted characters"
-            )
-        return True
+        issues.append(f"book Name metadata ({name}) does not meet naming conventions")
+
     title_length = len(regex.findall(r"\X", book.zim_metadata["Title"]))
 
     if title_length > Context.zim_title_max_length:
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Title metadata is {title_length} characters long"
-            )
-        return True
+        issues.append(
+            f"book Title metadata is {title_length} characters long, "
+            f"maximum length: {Context.zim_title_max_length}"
+        )
 
     description_length = len(regex.findall(r"\X", book.zim_metadata["Description"]))
     if description_length > Context.zim_description_max_length:
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Description metadata is {description_length} "
-                "characters long"
-            )
-        return True
+        issues.append(
+            f"book Description metadata is {description_length} characters long, "
+            f"maximum length: {Context.zim_description_max_length}"
+        )
     flavour = book.zim_metadata.get("Flavour")
     if flavour and not flavour.isalpha():
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Flavour metadata contains non-alphabetic characters"
-            )
+        issues.append(
+            f"book Flavour metadata ({flavour}) contains non-alphabetic characters"
+        )
 
-        return True
-    return False
+    return issues
 
 
-def update_book_issues(
-    session: OrmSession,
-    book: Book,
-    *,
-    update_events: bool = False,
-    raise_exceptions: bool = False,
-):
+def get_book_issues(
+    session: OrmSession, book: Book, *, raise_exceptions: bool = False
+) -> dict[str, list[str]]:
     """
-    Update book issues based on it's associated title and optionally update book events.
+    Compute book issues based on it's associated title
 
-    A book's issues will not be updated if:
-    - it has no associated title
-    - it is missing mandatory metadata keys
-    - it's associated title is missing mandatory metadata
-    - it's location kind is to_delete or deleted
+    Makes the same assumptions as the update_book_issues function
     """
-    if (
-        not book.title
-        or get_missing_metadata_keys(book.zim_metadata)
-        or title_is_missing_mandatory_metadata(book.title)
-        or book.location_kind in ["deleted", "to_delete"]
-    ):
-        return
-
-    issues: list[str] = []
-
+    issues: dict[str, list[str]] = {}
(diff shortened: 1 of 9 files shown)


backend/src/cms_backend/db/book.py
```diff
@@ -550,96 +550,60 @@ def revert_book(
...
     if different_metadata_keys:
-        issues.append("metadata mismatch")
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book metadata is different from title metadata: "
-                f"{','.join(different_metadata_keys)}"
-            )
+        issues["metadata mismatch"] = [
+            "book metadata is different from title metadata: "
+            f"{','.join(different_metadata_keys)}"
```
benoit74, line 617: We need to show metadata differences (value in book, value in title). I know it is a bit complex due to illustrations, but we can base64 encode values which are expected to be quite small. And have a tweak in UI to display images rather than base64 value when metadata is an illustration.

backend/src/cms_backend/db/book.py
```diff
@@ -550,96 +550,60 @@ def revert_book(
...
             unknown_languages.append(language_code)
 
     if unknown_languages:
-        issues.append("invalid language code")
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book has unknown language code(s) {unknown_languages}"
-            )
+        issues["invalid language code"] = [
+            f"book has unknown language code(s) {unknown_languages}"
```
benoit74: We should join values with a comma for better rendering

backend/src/cms_backend/db/book.py
```diff
@@ -550,96 +550,60 @@ def revert_book(
...
+        ]
 
-    if has_flavour_mismatch(book.flavour, book.title.flavours):
-        issues.append("flavour mismatch")
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book flavour is not in list of title flavours"
-            )
+    if has_flavour_mismatch(book.flavour, book.title.flavours):  # pyright: ignore[reportOptionalMemberAccess]
+        issues["flavour mismatch"] = ["book flavour is not in list of title flavours"]
```
benoit74: We should show book flavour and title flavours in the message

backend/src/cms_backend/db/book.py
```diff
@@ -550,96 +550,60 @@ def revert_book(
...
+
     title_length = len(regex.findall(r"\X", book.zim_metadata["Title"]))
 
     if title_length > Context.zim_title_max_length:
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Title metadata is {title_length} characters long"
-            )
-        return True
+        issues.append(f"book Title metadata is {title_length} characters long")
```
benoit74: Show `zim_title_max_length` in message for clarity

backend/src/cms_backend/db/book.py
```diff
@@ -550,96 +550,60 @@ def revert_book(
...
     description_length = len(regex.findall(r"\X", book.zim_metadata["Description"]))
     if description_length > Context.zim_description_max_length:
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Description metadata is {description_length} "
-                "characters long"
-            )
-        return True
+        issues.append(
+            f"book Description metadata is {description_length} characters long"
```
benoit74: Show `zim_description_max_length` in the message for clarity

backend/src/cms_backend/db/book.py
```diff
@@ -550,96 +550,60 @@ def revert_book(
...
+        )
     flavour = book.zim_metadata.get("Flavour")
     if flavour and not flavour.isalpha():
-        if update_events:
-            book.events.append(
-                f"{getnow()}: book Flavour metadata contains non-alphabetic characters"
-            )
+        issues.append("book Flavour metadata contains non-alphabetic characters")
```
benoit74: Show book flavour in message for clarity

backend/src/cms_backend/db/book.py
```diff
@@ -679,24 +643,62 @@ def update_book_issues(session: OrmSession, book: Book, *, update_events: bool =
...
+    if not can_compute_book_issues(book):
+        return
+
+    issues = get_book_issues(session, book)
+    book.issues = list(issues.keys())
+    if update_events:
+        now = getnow()
+        for reasons in issues.values():
+            for reason in reasons:
+                book.events.append(f"{now}: {reason}")
```
benoit74: This is going to be very wordy now, isn't it? I think we can save in events something summarized, like only the kind of issues. And maybe only one event. Something like `",".join(issues.keys())` is maybe sufficient

benoit74 (changes requested):
Thank you. 

Some changes required to make UI even more usable by CMS user, but I think your approach is indeed sufficient for the time being. I imagined something way more complex (like dedicated components for each kind of issues) but a list of strings is indeed probably sufficient.

Shall we add an "Edit Title" button as a shortcut to direct go to title edit tab? Ideally should be shown only when issues related to title are shown, but I'm not sure it is that easy to implement. If not, we can show it unconditionally.

backend/src/cms_backend/db/book.py
elfkuzco: I opted instead to pass the book's title which has already been loaded and used the existing `DiffViewer` component to render the differences. See the last image in updated PR description

elfkuzco:
Had to rebase to incorporate the zimcheck errors detection into this PR

backend/src/cms_backend/db/book.py
benoit74: Great!

benoit74 (approved):
LGTM

elfkuzco:
@benoit74 , i added a new commit just now. Doesn't affect anything other than returning issues list instead of raising an error that issues cannot be commuted. Rationale is that if we cannot compute issues, then, book likely doens't have any issues to start with and it's not particularly helpful  telling user "Book does not meet requirements for issues to be computed". WDYT?

benoit74:
Indeed, you are right, thanks

PR #365, opened 2026-06-25
show book title name instead of ID in book details

### LINKED ISSUE #361: Show title name instead of title id in book details
Currently we see the title ID:

Image

We should see the title name instead (even if technically speaking the link is to the ID).

elfkuzco:
## Changes
- return book title name in API response
- display book title name instead of ID in book detail view

This closes #361


benoit74 (approved):
LGTM

PR #366, opened 2026-06-25
filter inbox books by name instead of ID

### LINKED ISSUE #362: Filter by book name instead of ID in inbox
Currently, it is possible to filter books by ID in the inbox:

Image

This does not really makes sense, we should rather allow to filter by book name, just like in the books list.

elfkuzco:
# Changes
- filter inbox books by name instead of ID

Screenshot_20260625_100212
Screenshot_20260625_100159

This closes #362


benoit74 (approved):
LGTM

PR #367, opened 2026-06-25
reword text shown during title archive

### LINKED ISSUE #363: Title archive text is imprecise
Image

This is misleading because books will be marked for deletion.

`This will hide it from the main title list but preserve all data.` => `This will delete title books, hide title from the main title list, but preserve title configuration data in archive.`

elfkuzco:
Screenshot_20260625_100940

## Changes
- reword text shown during title archive

This closes #363


benoit74 (approved):
LGTM

PR #369, opened 2026-06-25
copy to temporary location before move

### LINKED ISSUE #368: Shuttle: moving file should be not create partially complete ZIM files
Currently, when the shuttle moves a file from `folderA/filename.zim` to `folderB/filename.zim`, it does:
- `cp folderA/filename.zim folderB/filename.zim` 
- `rm folderA/filename.zim`

This is problematic because it means that `folderB/filename.zim` appears while not yet complete.

We need to change the logic to introduce a temporary file so that we have an atomic `mv` operation to make ZIM file appear in target:
- `cp folderA/filename.zim folderB/filename.zim.tmp` (with `.tmp` suffix extension)
- `mv folderB/filename.zim.tmp folderB/filename.zim` 
- `rm folderA/filename.zim`

The rest of our tooling (will) take care of considering only `.zim` files.

elfkuzco:
## Changes
- during move operation, copy ZIM files to temporary location before making move

This fixes #368

### DIFF
--- a/backend/src/cms_backend/shuttle/move_files.py
+++ b/backend/src/cms_backend/shuttle/move_files.py
@@ -111,7 +111,11 @@ def move_book_files(session: OrmSession, book: Book):
             ShuttleContext.local_warehouse_paths
         )
         target_path.parent.mkdir(parents=True, exist_ok=True)
-        shutil.copy(source_path, target_path)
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM

PR #370, opened 2026-06-25
fix logic for detecting if title form is valid

elfkuzco:
## Changes
- apply title regex to name instead of title metdata
- fix logic for detecting if title metadata or name is valid


benoit74 (approved):
LGTM, nice catch

PR #375, opened 2026-06-25
omit processed books in shuttle/mill actions

### LINKED ISSUE #374: Logic to retrieve zimcheck results is flawed
Since the query now updates DB at each modification, there is no need to increment "skip" values or we skip too many records.

Please check if we do not have another place with same problem.

elfkuzco:
## Changes
- omit processed books as shuttle/mill loops around to find books to process instead of incrementing skip

This fixes #374


benoit74 (changes requested):
Do we really need the `skip` variable if it is always `0`?

benoit74 (approved):
LGTM

PR #378, opened 2026-06-26
compare images by hash

elfkuzco:
## Changes
- compute illustration hashes and store in title, book and title history
- use hash value as comparison when comparing metadata keys
- add maint-script to compute hashes of existing title, book and title history
- compute book/title hash during creation
- update illustration hash during title upate
- use hashes to detect whether to show illustrations as differences


elfkuzco:
closing this in favour of #383

PR #380, opened 2026-06-26
expose current state of zimcheck result url in get_zim_urls

### LINKED ISSUE #377: Supply Zimcheck URL to the outside world (zimfarm for now)
Since CMS now has the ultimate responsibility to decide when zimcheck JSON file is deleted from S3, the CMS should publish this information. We can even imagine CMS might provide enhanced versions, recompute the zimcheck, etc in the future.

I propose to simply add a new kind of link to `GET /books/zims` endpoint, and populate it with zimcheck URL when zimcheck has not been deleted from S3.

elfkuzco:
## Changes
- add zimcheck url to result of get_zim_urls if book has a zimcheck url that has not been deleted
- update ui component for showing zimurl buttons to omit urls of kind `zimcheck`

This  closes #377

### DIFF
--- a/backend/src/cms_backend/db/books.py
+++ b/backend/src/cms_backend/db/books.py
@@ -197,6 +197,8 @@ def get_zim_urls_prod(session: OrmSession, zim_ids: list[UUID]) -> ZimUrlsSchema
     stmt = (
         select(
             Book.id.label("book_id"),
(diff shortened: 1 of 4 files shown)


benoit74 (approved):
LGTM

PR #384, opened 2026-06-26
update book issues when issues are fetched from API

### LINKED ISSUE #381: Save recomputed issues when book issues are fetched from the API
Whenever book issues are fetched from the API, we should save results in the DB. It is understood that constraints will change over time and it is hence normal that the "summary" stored in book DB table becomes "rotted", but it is too hard to understand for the user that UI displays there is no issues in "issues" panel while there are still issues in other panels.

This means we have write operations occurring on a GET endpoint. This is acceptable, but I recommend we then restrict the endpoint to users with "update" permissions on books. UI should hence not show the "issues" tab unless user has this permission. This matches current users roles, and avoids opening a weird "write without permission" hole.

elfkuzco:
## Changes
- restrict API/UI to view book issues to accounts with `update` permissions on `book`
- update book issues when API to fetch book issues is called

This closes #381


benoit74 (approved):
LGTM

PR #385, opened 2026-06-29
show illustration file sizes in bytes

elfkuzco:
## Changes
- show illustration file sizes in bytes both in title form and inline image editor

Screenshot_20260629_124230
Screenshot_20260629_124215


benoit74 (approved):
LGTM

PR #389, opened 2026-07-01
add increase/decrease thresholds for article/media count

### LINKED ISSUE #387: Article and media count should have less strict limit on increase than decrease
We need a less strict limit on increase of article or media limit than on decrease. We can start by saying that increase acceptable ratio is twice decrease acceptable ratio.

elfkuzco:
## Changes
- add different variables for article/media count change thresholds. New variables control different change directions (decrease or increase) as opposed to one for controlling the change
- add migration script to populate variables based on existing change values
- move logic to get media/article count issues to dedicated function

This closes #387


backend/src/cms_backend/db/book.py
```diff
@@ -595,6 +595,102 @@ def get_book_metadata_issues(book: Book) -> list[str]:
     return issues
 
 
+def get_book_article_count_issues(*, book: Book, latest_book: Book) -> list[str]:
```
benoit74, line 598: Please display the threshold in message, so that it is even clearer why we have the issue (and how "bad" the issue is)

benoit74 (approved):
One very small remark, you can fix it on your own and merge.

Issue #333, opened 2026-06-09
Better solve book metadata issue


benoit74:
Currently, when a book has a metadata issue, we "only" allows to fix the issue by going to title edit screen, which is indeed the most convenient place to do so.

Just like in #255, we should have a dedicated screen to solve this metadata issue, accessible directly from the book details screen.

Once #255 will be fixed, we should also show Zimfarm recipe(s) metadata so that user has a better understanding of the situation: which metadata the book has, which metadata is configured on the Zimfarm, which metadata is configured on the title.

When title metadata is updated (through this new screen), CMS should also propose to update Zimfarm recipe(s) at the same time, assuming the user has sufficient rights.

PR #390, opened 2026-07-01
filter books by offliner or issues and show books as cards

### LINKED ISSUE #388: Book inbox should display scraper and allow to filter by scraper and book issue type
On book inbox, we need to:
- display scraper
- better display list of issues (it is too "hidden" ATM)
- filter by scraper
- filter by book issue type

Regarding display, I wonder if it is not the right time to move to card display rather than table, especially since we do not intend to allow to sort columns.

elfkuzco:
Screenshot_20260701_210236

## Changes
- make param to filter books a pydantic model
- filter books by issue or scraper name
- add UI filters to filter books by issue and/or scraper name
- make inbox book table to always display as card instead of normal table since it has so many columns

This closes #388


elfkuzco:
@benoit74 , there are other places where the BookTable is used. For example in books view and in title detail view. Should we make those cards too? And should we extend the filters to teh normal books view?

benoit74 (changes requested):
Thank you

- Seeing it in action, I don't feel like displaying the whole scraper metadata and allowing to search inside it in free text is the good approach. I suggest we rather replace (in the UI) the "Scraper" filter with an "Offliner" filter which is a list of offliners as currently defined in Zimfarm (I suggest UI call Zimfarm API to retrieve list of offliners). 
- The UI should then not show the "Scraper" metadata in the card, but the "Offliner" (just matching the scraper metadata, and saying "Unknown" if there is no match)
- The card width can then be made much smaller, with at least 3 or maybe even 4 cards side-by-side
- The card UI should be polished: all "lines" should have the same height (except the list of issues where we want one issue per "line" + one line for the "view issues" button), data should not overflow on two lines but be wrapped with ellipsis, elements of "status" should be vertically aligned.

elfkuzco:
> I suggest we rather replace (in the UI) the "Scraper" filter with an "Offliner" filter which is a list of offliners as currently defined in Zimfarm (I suggest UI call Zimfarm API to retrieve list of offliners).

This will require zimfarm to whitelist cms UI due to CORS

benoit74:
> This will require zimfarm to whitelist cms UI due to CORS

OK, we will need to do this anyway for #333

elfkuzco:
> Seeing it in action, I don't feel like displaying the whole scraper metadata and allowing to search inside it in free text is the good approach. I suggest we rather replace (in the UI) the "Scraper" filter with an "Offliner" filter which is a list of offliners as currently defined in Zimfarm (I suggest UI call Zimfarm API to retrieve list of offliners).

Should we also show "unknown" when user doesn't pass any offliner? Also, the book full details inherits this attribute, should it also show "Unknown" instead of the underlying "scraper"?

elfkuzco:
Screenshot_20260702_133818

benoit74:
Thank you.

- I see that "Status" line is still higher than other lines. While I do understand "Issues" line must be de-facto higher when we have multiple issues, I would prefer we have the same height on all lines when there is no issue. Probably worth to set a min-height on lines so that "Status" line content "fits" inside.

- I feel the "View Issues" button very confusing, it is weird that clicking anywhere on the card brings you to details but clicking on this specific area brings you to Issues. Please rather add two buttons at the bottom of the card, only to view details and one to view issues (not show when there is no issues), and make the card not clickable.

- Flavour is the only field to not have a "dash" when there is no value ; this is weird, please put a "dash" when there is no flavour set

All that being said, looking back at Zimfarm files card, I feel like we should adopt a very similar design for consistency and because I like it: 
- no unneeded "field name" repeated on all cards when not needed (we probably never need it on CMS, see suggestions below)
- no text on buttons, make icon self-explanatory + tooltip

image

This could even solve the issue of "Status line" height ... by not have a status line anymore, but just displaying the status since it is kinda obvious it is the status. Same for issues, the fact it is a badge in red make it quite clear it is an issue. Deletion date could be shown just like "date" but with an different and explicit icon. Offliner could also have its own badge with an explicit icon (everyone will know it is the offliner when value text will be correct anyway). And since buttons will have no text, we can add the "view" + "download" buttons directly on the card just like on ZF which is a big plus.

> Should we also show "unknown" when user doesn't pass any offliner? Also, the book full details inherits this attribute, should it also show "Unknown" instead of the underlying "scraper"?

We should compute offliner based on scraper value + matching with the current list of known offliner. So "Unknown" should not be shown in your screenshot, we should see "mwoffliner" (and maybe "zimit", I don't recall ^^).

elfkuzco:
I have applied the suggestions and this is what it looks like now:

Screenshot_20260706_122603
Screenshot_20260706_122540


The second image is from the title detail view. I opted to instead break the original status component into a seperate status and location component. Card header would become crowded if I add Status information to it.

benoit74 (changes requested):
Please change icons + their order so that most likely icons are first (left) and so that we have same icons than on Zimfarm:
- first book details with `mdi-information` icon
- then view / download icons (this is OK)
- then book issues with `mdi-bell-alert` icon

benoit74 (approved):
LGTM

PR #392, opened 2026-07-02
Fix title name regex: language might need 3 characters

benoit74:
E.g. `ted_mul_audacious-project` is currently invalid while it is valid


benoit74:
@elfkuzco FYI

PR #393, opened 2026-07-02
change wording for change threshold

### LINKED ISSUE #391: Change treshold displayed is wrong
Image

Should be 20% instead of 0.2%.

And "change threshold" is not really clear, I think we can find something more explicit.

elfkuzco:
## Changes
- multiply threshold setting by 100 for display

This fixes #391


benoit74 (approved):
LGTM

PR #405, opened 2026-07-03
add API/UI logic to promote book

### LINKED ISSUE #395: Implement book validation operation
Currently, when CMS editor want to move a book forward in the pipeline (from quarantine to staging or production, or from staging to production), he needs to perform actions in various places depending on the issue(s) book has in quarantine or in staging. This is not convenient at all for the CMS user.

We need to implement an atomic "validate" operation on books.

Book validation is possible only if book is in quarantine or staging.

What validation entails will depend widely on what is currently happening to the book. It could means CMS will have to create a title, alter existing title properties (add a flavour, change metadata, ...), move book to a different location, etc ... Some of these changes are required (e.g. it is not possible to choose if we add book flavour to title) while other are optional (e.g CMS user can choose to not update title metadata). 

Because of these, we need to have a dry-run mode where CMS API will only return the list of actions expected to be done, with their kind and data ; e.g. `{"kind": "create_title", "required": true, "data": { "title_name": "test_en_all"}}`. In "apply" mode, the API will receive the list of actions to run and the list of actions to ignore. Before applying, it will compute again actions to perform again ; if list is not identical, then it will fail assume some things have changed between dry-run and apply, asking for a new validation.

In the UI, when user choose to validate a book, results of the dry-run will be presented. By default all actions will be checked, and user can uncheck optional actions he doesn't want to run.

In "apply", some action will accept to have additional inputs from the user. E.g. the `create_title` action accepts a list of collection+path to associate the title with. UI will allow to set these in the validation for maximum convenience.

@elfkuzco can you check if all this make sense? Can you prepare the list of all actions types we currently have to support (e.g. list of issues we might want to solve to move from staging to prod, etc ...), and whether all this make sense or not?

elfkuzco:
Screenshot_20260705_153920
Screenshot_20260705_152543
Screenshot_20260705_152459

## Changes
- add functions to compute actions needed to move book from current location to prod
- add function to apply selected actions needed to move book from current location to prod

This closes #395

### DIFF
--- a/backend/src/cms_backend/db/book_actions.py
+++ b/backend/src/cms_backend/db/book_actions.py
@@ -1,0 +1,307 @@
+from collections import deque
+from typing import Any
+from uuid import UUID
+
+from sqlalchemy.orm import Session as OrmSession
+
+from cms_backend.db.book import (
+    get_book,
+    get_book_or_none,
+    get_book_unsupported_languages,
+    get_differing_metadata_keys,
+    get_zimcheck_errors,
+    move_book_to_destination,
+)
+from cms_backend.db.exceptions import RecordDoesNotExistError
+from cms_backend.db.rules import (
+    has_flavour_mismatch,
+    title_is_missing_mandatory_metadata,
+)
+from cms_backend.db.title import create_title, restore_title, update_title
+from cms_backend.schemas.models import (
+    BaseBookPromotionAction,
+    BookPromotionAction,
+    RestoreTitlesSchema,
+    TitleCreateSchema,
+    TitleUpdateSchema,
+)
+from cms_backend.schemas.orms import (
+    ZimcheckSummarySchema,
+)
+from cms_backend.utils.zim import (
+    get_missing_keys,
+    get_missing_metadata_keys,
+)
+
+
+def get_book_promotion_actions(
+    session: OrmSession, *, book_id: UUID
+) -> list[BookPromotionAction]:
+    """Get actions required to promote a book to 'prod'."""
+    book = get_book_or_none(
+        session,
+        book_id=book_id,
+        has_error=False,
+        needs_file_operation=False,
+        needs_processing=False,
+        locations=["staging", "quarantine"],
+    )
+    if book is None:
+        raise RecordDoesNotExistError(
+            f"Book {book_id} does not meet criteria to be validated"
+        )
+
+    actions: list[BookPromotionAction] = []
+
+    missing_metadata_keys = get_missing_metadata_keys(book.zim_metadata)
+    if missing_metadata_keys:
+        raise ValueError(
+            "Book is missing mandatory metadata keys and cannot "
+            "possibly be promoted through to 'prod'"
+        )
+
+    unknown_languages = get_book_unsupported_languages(book)
+    if unknown_languages:
+        actions.append(
+            BookPromotionAction(
+                kind="unknown_languages",
+                requirement="information",
+                data={},
+                message=(
+                    "Book has unknown language code(s): "
+                    f"{','.join(unknown_languages)}. "
+                    "Please contact a CMS admin if code(s) look legit to you."
+                ),
+            )
+        )
+
+    zimcheck_errors = get_zimcheck_errors(book, raise_exceptions=False)
+    if zimcheck_errors and book.zimcheck_summary:
+        zimcheck_summary = ZimcheckSummarySchema.model_validate(book.zimcheck_summary)
+        actions.append(
+            BookPromotionAction(
+                kind="zimcheck_issues",
+                requirement="information",
+                data={},
+                message=f"Book has {zimcheck_summary.error_count} zimcheck error(s). ",
+            )
+        )
+
+    if book.title is None:
+        actions.append(
+            BookPromotionAction(
+                kind="create_title",
+                requirement="mandatory",
+                data={
+                    "name": book.name,
+                    "maturity": "stable",
+                    "title": book.zim_metadata["Title"],
+                    "creator": book.zim_metadata["Creator"],
+                    "publisher": book.zim_metadata["Publisher"],
+                    "description": book.zim_metadata["Description"],
+                    "language": book.zim_metadata["Language"],
+                    "illustration_48x48_at_1": book.zim_metadata[
+                        "Illustration_48x48@1"
+                    ],
+                    "flavours": [book.flavour],
+                    "collection_titles": [],
+                },
+                message="Create new title. Please configure title collection(s).",
+            )
+        )
+        return actions
+
+    title = book.title
+
+    if title.archived:
+        actions.append(
+            BookPromotionAction(
+                kind="restore_title",
+                requirement="mandatory",
+                data={"title_names": [title.name]},
+                message=f"Restore title '{title.name}' from archive",
+            )
+        )
+
+    if title_is_missing_mandatory_metadata(title) or get_differing_metadata_keys(book):
+        actions.append(
+            BookPromotionAction(
+                kind="update_title_metadata",
+                requirement="optional",
+                data={
+                    "title": book.zim_metadata["Title"],
+                    "creator": book.zim_metadata["Creator"],
+                    "publisher": book.zim_metadata["Publisher"],
+                    "description": book.zim_metadata["Description"],
+                    "language": book.zim_metadata["Language"],
+                    "illustration_48x48_at_1": book.zim_metadata[
+                        "Illustration_48x48@1"
+                    ],
+                },
+                message="Update title metadata",
+            )
+        )
+
+    if title.maturity != "stable":
+        actions.append(
+            BookPromotionAction(
+                kind="update_title_maturity",
+                requirement="optional",
+                data={
+                    "maturity": "stable",
+                },
+                message=f"Mark title '{title.maturity}' as stable.",
+            )
+        )
+    if len(title.collections) == 0:
+        actions.append(
+            BookPromotionAction(
+                kind="set_title_collections",
+                requirement="mandatory",
+                data={
+                    "collection_titles": [],
+                },
+                message="Configure title collection(s)",
+            )
+        )
+    if has_flavour_mismatch(book.flavour, title.flavours):
+        actions.append(
+            BookPromotionAction(
+                kind="update_title_flavours",
+                requirement="mandatory",
+                data={"flavours": [book.flavour, *title.flavours]},
+                message=(
+                    f"Add '{book.flavour}' to title flavours: "
+                    f"{','.join(title.flavours)}"
+                ),
+            )
+        )
+
+    return actions
+
+
+def apply_book_promotion_actions(
+    session: OrmSession,
+    *,
+    book_id: UUID,
+    actions: list[BaseBookPromotionAction],
+    author_id: UUID,
+) -> None:
+    """Apply a list of actions to book so that it can be promoted to 'prod'"""
+    if len(actions) == 0:
+        raise ValueError("At least one action must be provided to promote the book")
+    action_kinds = {action.kind for action in actions}
+    if len(action_kinds) != len(actions):
+        raise ValueError("Provided actions contain duplicates")
+
+    book = get_book(session, book_id=book_id)
+    expected_actions = get_book_promotion_actions(session, book_id=book_id)
+
+    expected_actions_set = {action.kind for action in expected_actions}
+    provided_actions_set = {action.kind for action in actions}
+
+    unknown_actions = provided_actions_set - expected_actions_set
+    if unknown_actions:
+        raise ValueError(
+            "One or more provided actions are not in the list of expected actions "
+            f"to promote the book. Unexpected actions: "
+            f"{','.join(sorted(unknown_actions))}; "
+            f"expected actions: {','.join(sorted(expected_actions_set))}"
+        )
+    expected_mandatory_actions = {
+        action.kind for action in expected_actions if action.requirement == "mandatory"
+    }
+    provided_mandatory_actions = {
+        action.kind for action in actions if action.requirement == "mandatory"
+    }
+    if expected_mandatory_actions != provided_mandatory_actions:
+        raise ValueError(
+            "The mandatory actions provided do not match the expected mandatory "
+            "actions. Expected mandatory actions: "
+            f"{','.join(sorted(expected_mandatory_actions))}; provided mandatory "
+            f"actions: {','.join(sorted(provided_mandatory_actions))}"
+        )
+    actions_todo = deque(actions)
+    # Batch the updates to a title in a common payload
+    title_update_payload: dict[str, Any] = {}
+    while actions_todo:
+        action = actions_todo.popleft()
+        match action.kind:
+            case "create_title":
+                payload = TitleCreateSchema.model_validate(action.data)
+                if not payload.collection_titles:
+                    raise ValueError(
+                        "Title must have at least one collection configured."
+                    )
+                book.title = create_title(session, author_id=author_id, payload=payload)
+            case "restore_title":
+                if not book.title:
+                    raise ValueError(
+                        "Book does not have an associated title to restore"
+                    )
+                payload = RestoreTitlesSchema.model_validate(action.data)
+                if (
+                    len(payload.title_names) != 1
+                    or payload.title_names[0] != book.title.name
+                ):
+                    raise ValueError(
+                        "Only the book's title should be specified in "
+                        "the restore payload"
+                    )
+                book.title = restore_title(
+                    session,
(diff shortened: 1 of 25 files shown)


elfkuzco:
Even after applying actions, i'm not sure how to ignore some issues because even though books are re-processed, it sill flows through `process_book`. Do we want to ignore some issues like `media_count`, etc?

elfkuzco:
@benoit74 , I have added the UI components to show text input for the actions compted by the backend and updated PR description

benoit74:
> Do we want to ignore some issues like `media_count`, etc?

We should not *ignore* any issue, but these issues will indeed never be fixed and will stay "as-is" once book is promoted.

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+    missing_metadata_keys = get_missing_metadata_keys(book.zim_metadata)
+    if missing_metadata_keys:
+        raise ValueError(
+            "Book is missing mandatory metadata keys and cannot "
+            "possibly be promoted through to 'prod'"
+        )
+
+    metadata_issues = get_book_metadata_issues(book)
+    if metadata_issues:
+        raise ValueError(
```
benoit74: This should not raise an error, but rather simply display the same kind of message than unknown languages saying bad metadata have been found.

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+    if unknown_languages:
+        actions.append(
+            BookPromotionAction(
+                kind="whitelist_language_codes",
+                required=False,
+                data={},
+                message=(
+                    "Book has unsupported language code(s): "
+                    f"{','.join(unknown_languages)}. "
+                    "Please contact a CMS admin to whitelist these codes."
```
benoit74: `Please contact a CMS admin to whitelist these codes.` => `Please contact a CMS admin if code(s) look legit to you.`

backend/src/cms_backend/schemas/models.py
```diff
@@ -124,6 +124,32 @@ class TitleUpdateSchema(BaseTitleCreateUpdateSchema):
...
+    "update_title_maturity",
+    "update_title_collections",
+    "update_title_flavours",
+]
+
+
+class BaseBookPromotionAction(BaseModel):
+    kind: BookPromotionActionKind
+    data: dict[str, Any]
+    required: bool
```
benoit74: Having to mark actions without impact (like `whitelist_language_codes`) as required "False/True" is complex because both are misleading the user.

I feel like we rather need a tri-state property: "information", "optional", "mandatory".

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+            "Book has bad metadata and cannot be promoted through to 'prod'.\n"
+            f"{'\n'.join(metadata_issues)}"
+        )
+
+    actions: list[BookPromotionAction] = []
+    unknown_languages = get_book_unsupported_languages(book)
+    if unknown_languages:
+        actions.append(
+            BookPromotionAction(
+                kind="whitelist_language_codes",
```
benoit74: `whitelist_language_codes` => `information_unknown_languages`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                ),
+            )
+        )
+
+    zimcheck_errors = get_zimcheck_errors(book, raise_exceptions=False)
+    if zimcheck_errors and book.zimcheck_summary:
+        zimcheck_summary = ZimcheckSummarySchema.model_validate(book.zimcheck_summary)
+        actions.append(
+            BookPromotionAction(
+                kind="whitelist_scraper_from_zimcheck",
```
benoit74: `whitelist_scraper_from_zimcheck` => `information_zimcheck_issues`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                required=False,
+                data={},
+                message=(
+                    f"Book has {zimcheck_summary.error_count} zimcheck error(s). "
+                    "Please contact a CMS admin to whitelist this scraper "
```
benoit74: Do not propose to contact admins, this is not "normal" CMS behavior to have to whitelist a scraper, more an interim solution until all scrapers are OK.

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+    actions: list[BookPromotionAction] = []
+    unknown_languages = get_book_unsupported_languages(book)
+    if unknown_languages:
+        actions.append(
+            BookPromotionAction(
+                kind="whitelist_language_codes",
+                required=False,
+                data={},
+                message=(
+                    "Book has unsupported language code(s): "
```
benoit74: unsupported => unknown

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                    "title": book.zim_metadata["Title"],
+                    "creator": book.zim_metadata["Creator"],
+                    "publisher": book.zim_metadata["Publisher"],
+                    "description": book.zim_metadata["Description"],
+                    "language": book.zim_metadata["Language"],
+                    "illustration_48x48_at_1": book.zim_metadata[
+                        "Illustration_48x48@1"
+                    ],
+                    "flavours": [] if book.flavour is None else [book.flavour],
+                    "collection_titles": [{"collection_name": "", "path": ""}],
```
benoit74: I don't get why we add an empty collection ... I would rather simply let title without any collection at this stage.

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                },
+                message=(
+                    "No title exists for this book. "
+                    "Create one from the book's metadata. "
+                    "Be sure to add collection entries so the book is assigned to "
```
benoit74: `Be sure to add collection entries so the book is assigned to a location` => `Please configure title collection(s).`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                    "description": book.zim_metadata["Description"],
+                    "language": book.zim_metadata["Language"],
+                    "illustration_48x48_at_1": book.zim_metadata[
+                        "Illustration_48x48@1"
+                    ],
+                    "flavours": [] if book.flavour is None else [book.flavour],
+                    "collection_titles": [{"collection_name": "", "path": ""}],
+                },
+                message=(
+                    "No title exists for this book. "
```
benoit74: `No title exists for this book. Create one from the book's metadata` => `Create new title`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                    "creator": book.zim_metadata["Creator"],
+                    "publisher": book.zim_metadata["Publisher"],
+                    "description": book.zim_metadata["Description"],
+                    "language": book.zim_metadata["Language"],
+                    "illustration_48x48_at_1": book.zim_metadata[
+                        "Illustration_48x48@1"
+                    ],
+                },
+                message=(
+                    "Title is missing or has outdated mandatory metadata. "
```
benoit74: `Update title metadata`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                data={"title_names": [title.name]},
+                message=f"Restore title '{title.name}' from archive",
+            )
+        )
+
+    if title_is_missing_mandatory_metadata(title) or get_differing_metadata_keys(book):
+        actions.append(
+            BookPromotionAction(
+                kind="update_title_metadata",
+                required=True,
```
benoit74: I don't think it is required, we should be allowed to keep existing title metadata if book ones are wrong

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+    if title.maturity != "stable":
+        actions.append(
+            BookPromotionAction(
+                kind="update_title_maturity",
+                required=False,
+                data={
+                    "maturity": "stable",
+                },
+                message=(
+                    f"Title maturity is '{title.maturity}', which prevents "
```
benoit74: => `Mark title '{title.maturity}' as stable`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                    "books from being promoted directly to 'prod'. "
+                    "Update it to 'stable'."
+                ),
+            )
+        )
+    if len(title.collections) == 0:
+        actions.append(
+            BookPromotionAction(
+                kind="update_title_collections",
```
benoit74: update_title_collections => set_title_collections

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                required=True,
+                data={
+                    "collection_titles": [{"collection_name": "", "path": ""}],
```
benoit74: Why having at least a fake collection?

elfkuzco:
> We should not ignore any issue, but these issues will indeed never be fixed and will stay "as-is" once book is promoted.

Okay. But we could apply promotion actions and book would still end up in staging because it has issues (either from zimcheck) in which case user would have to manually move to prod. Should we pass settings to create_title_modified event which gets drilled down to the logic to determine if book goes to staging?

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+                },
+                message=(
+                    "Title has no collection entries. "
```
benoit74: => `Configure title collection(s)`

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,351 @@
...
+    create_title_updated_event = False
+    # Batch the updates to a title in a common payload
+    title_update_payload: dict[str, Any] = {}
+    while actions_todo:
+        action = actions_todo.popleft()
+        match action.kind:
+            case "create_title":
+                payload = TitleCreateSchema.model_validate(action.data)
+                if not payload.collection_titles:
+                    raise ValueError("Titles must be created with collection details")
```
benoit74: `Titles must be created with collection details` => `Title must have a least one collection configured`

benoit74 (changes requested):
Thank you, we are clearly in the proper direction.

I focused only on the backend so far for review, please fix that (in both backend and frontend if needed).

Please note that this PR made me realize we have another issue you might want to tackle at the same time: https://github.com/openzim/cms/issues/407

backend/src/cms_backend/db/book_actions.py
elfkuzco: It's more of a placeholder in case user wants to apply the data without going through the UI and it would free them from manually having to add the property

benoit74: OK got it, please add a comment for "future self" then, pretty hard to get otherwise.

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,348 @@
...
+
+    metadata_issues = get_book_metadata_issues(book)
+    if metadata_issues:
+        actions.append(
+            BookPromotionAction(
+                kind="metadata_issues",
+                requirement="information",
+                data={},
+                message=(
+                    "Book has bad metadata and cannot be promoted through to 'prod'.\n"
```
benoit74: Why can't we promote the book to "prod" if it has metadata issues? We should only let user fix these metadata at title level.

frontend/src/components/PromoteBookDialog.vue
```diff
@@ -0,0 +1,529 @@
...
+                    class="flex-grow-0"
+                    @click.stop="toggleAction(index)"
+                    @update:model-value="
+                      (v) => {
+                        if (action.requirement === 'optional' && v !== null)
+                          actionChecked[index] = v
+                      }
+                    "
+                  />
+                  <v-icon
```
benoit74, line 60: I propose to rather use "warning" color + mdi-alert-circle icons for "mandatory" + "optional", and "info" color + icon for "information" ; optional changes are still important to analyze and act upon

benoit74 (changes requested):
Some UI feedback:
- in promotion dry run results, I propose to uncolapse all sections by default so that user see what he has to validate /configure
- when creating a title, for some reason there is already a collection_title configured ("Kiwix/devdocs") ; this is wrong especially since if I try to promote there is no collection_title configured (which is OK, only the UI should not display a collection_title by default which is not really configured) ; in turn, this tricks the user which sees a message like "collection name is empty" which is very misleading ; either we fix this in API response (no collection title in response by default) or at UI level (drop "dummy" collection title when receiving dry-run results)
- for all update_xxx actions, we need to see current and future values, so that user can take an informed decision about what he is changing

backend/src/cms_backend/db/book_actions.py
elfkuzco: There will be no data for them to submit primarily because we don't update anything regarding book and updating title would enforce those constraints.

elfkuzco: if this is the only thing preventing a book from going to prod, then, the user can move the book directly. Something I need to point out is that promoting a book doesn't guarantee book will move to prod though because book will still have issues. Should I add additional logic to make sure book really goes to prod?

benoit74: > There will be no data for them to submit primarily because we don't update anything regarding book and updating title would enforce those constraints.

Indeed. Then I'm wondering, shouldn't we simply drop this action, which is somehow a duplicate of "create_title" and "update_title" actions anyway?

> if this is the only thing preventing a book from going to prod, then, the user can move the book directly. Something I need to point out is that promoting a book doesn't guarantee book will move to prod though because book will still have issues. Should I add additional logic to make sure book really goes to prod?

This is indeed a very good point, thank you for mentioning this, I clearly oversight this. We indeed need to ensure book is really moving to prod or fail the promotion.

I've been wondering if the "title_modified" event we've introduced is not just over engineering. Looking at real production cases, the event is always processed very fast. So all this async stuff seemed a bit useless to me since few weeks.

We now have a real case where in addition to been useless, it really hurts the user experience and the processing logic.

Would it help if we remove all this "async title modified" logic and process this title modification in same API call?

elfkuzco: > I've been wondering if the "title_modified" event we've introduced is not just over engineering. Looking at real production cases, the event is always processed very fast. So all this async stuff seemed a bit useless to me since few weeks.

I think it could still be useful because it processes all books belonging to a title. Maybe I've been "abusing" it and using it instead of processing a single book in this case. Obviously, with retention rules being applied, there might not be so many books it handles.

Plus, it is also used when new title is created

elfkuzco: I think, in this case, we can remove it entirely and call the logic to `move_book` with `prod` as destination since it only acts on a book

benoit74: Yep ; please remove it entirely, including UI screens ; sorry to realize this a bit late

elfkuzco: sorry, what do you mean by UI screens?

benoit74: I think we do not need anymore the "Inbox Events", at least we should not show it anymore. Or are we using for other kind of events?

And we should have an issue to drop DB table and API once we've confirmed it was the right choice (or we can drop it right now).

backend/src/cms_backend/db/book_actions.py
```diff
@@ -0,0 +1,346 @@
...
+        payload = TitleUpdateSchema.model_validate(title_update_payload)
+        book.title = update_title(
+            session,
+            title_identifier=book.title.name,
+            author_id=author_id,
+            payload=payload,
+            create_event=False,
+        )
+
+    current_location = next(
```
benoit74: Why not calling `move_book` directly? This looks a lot like duplicated code, and I have no clue what is really different. Maybe there is too much difference, but you need to explain these.

elfkuzco: `move_book` only requires books that are in staging/prod and promote book accepts books that are in quarantine.

elfkuzco: refactored the logic to `move_book` to use a helper function that is now reusable.

benoit74 (approved):
LGTM, thank you

PR #409, opened 2026-07-06
exclude to_delete books that are not being processed from pending operations

elfkuzco:
## Changes
- exclude to_delete books that don't need processing from hasPendingOperations UI check. This way, we can still see the button to Recover Book. to_delete books always have file operation set to True.


benoit74 (approved):
Nice catch!

PR #411, opened 2026-07-06
allow empty flavours

### LINKED ISSUE #407: Title should be allowed to have book with and without flavours
Given current modelling, it is not possible to have a book without flavour and a book with flavour in the same title. This should be possible. Note that it does already exists in the wild and is legit, see [mdwiki_en_all](https://cms.openzim.org/title/mdwiki_en_all).

I feel like we should add `""` (empty string) to the list of title flavours in DB for this to work.

elfkuzco:
## Changes
- allow flavour value sent via create/update for title to be empty string
- allow flavour value for book update to be empty string
- show empty option with label `Empty`

Screenshot_20260706_161602
Screenshot_20260706_161520
Screenshot_20260706_161504

This closes #407

### DIFF
--- a/backend/src/cms_backend/api/routes/fields.py
+++ b/backend/src/cms_backend/api/routes/fields.py
@@ -124,7 +124,7 @@ LangCode = Annotated[
 ]
 
 ZimFlavour = Annotated[
-    NotEmptyString,
+    str,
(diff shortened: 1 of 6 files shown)


benoit74 (approved):
LGTM

PR #413, opened 2026-07-06
display books list view as cards

### LINKED ISSUE #408: use book cards across book list and backup list views
https://github.com/openzim/cms/pull/390 introduced presenting books as cards in the inbox view. This new presentation mode should be extended to the book list views and backup list views

elfkuzco:
## Changes
- load zim urls of book from books base view
- display books as cards in book table

This closes #408


benoit74 (approved):
LGTM

I think that at some point we will need to drop display-mode + show-urls options on BookTable component since all components are doing the same rendering now ... but I'm ok to keep it until we are sure we do not want old table display mode.

PR #415, opened 2026-07-07
create book with empty flavour and make flavour non-nullable

elfkuzco:
## Changes
- create book with empty flavour instead of None when flavour is empty


benoit74 (commented):
Do I get it right it means we need to fix DB for all existing books + titles? 

And then shouldn't we make the book flavour column not nullable?

And why `zim_metadata.get("Flavour", None)` rather than `zim_metadata.get("Flavour", "")` (the later is significantly more obvious to read).

Finally, I've found many cases where we had a fallback logic should flavour be None ... if it is not possible anymore, then we should remove it for maintainability

elfkuzco:
> Do I get it right it means we need to fix DB for all existing books + titles?

Yeah, we will need a maint-script

> And then shouldn't we make the book flavour column not nullable?

> And why zim_metadata.get("Flavour", None) rather than zim_metadata.get("Flavour", "") (the later is significantly more obvious to read).

I don't know if there are books that will not contain the `Flavour` metadata given it's not one of the mandatory metadata. But there are quite some books that set it to empty string.

benoit74:
> I don't know if there are books that will not contain the Flavour metadata given it's not one of the mandatory metadata. But there are quite some books that set it to empty string.

We can assume missing Flavour metadata and empty Flavour metadata value are the same, no need to make a distinction in processing logic

elfkuzco:
Created a migration to make book/book_history flavour non-nullable and set those that are NULL to empty string

benoit74 (approved):

PR #416, opened 2026-07-07
hide metadata section when creating new title

### LINKED ISSUE #406: Do not propose to fill metadata in title creation popup
My understanding is that these fields are not to be filled, as metadata is taken from the zimfarm recipe. I find it a bit confusing, therefore, that those field are shown when creating a title. Or am I missing something and they _should_ be filled at creation?

Image

elfkuzco:
## Changes
- hide metadata section when creating new title

This closes #406

### DIFF
--- a/frontend/src/components/TitleForm.vue
+++ b/frontend/src/components/TitleForm.vue
@@ -60,10 +60,10 @@
       </v-row>
     </div>
 
-    <v-divider class="my-6" />
+    <v-divider v-if="isEditMode" class="my-6" />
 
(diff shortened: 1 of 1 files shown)


benoit74 (changes requested):
I don't get the relation between this PR content and this content. This looks like changes which should have been done in https://github.com/openzim/cms/pull/415

elfkuzco:
PR branch automatically changed base. Last commit is the only change. Will rebase.

benoit74 (approved):
LGTM

PR #418, opened 2026-07-08
prioritize local auth and redirect after sign in

### LINKED ISSUE #414: CMS UI interactions with Zimfarm should be polished
I see two issues:
- when Zimfarm is not responding appropriately, we need to handle the error nicely and have a proper "degraded" mode ; for now the UI just breaks ; this is important especially since we could have downtime on Zimfarm and it should not propagate to CMS UI
- there is something to polish around local auth ; not sure if it is at CMS and/or Zimfarm level, but I'm seeing calls to retrieve Zimfarm failing + calls to ory staging when running locally ; do we even have something in place to support CMS + Zimfarm local auth scenarii?

elfkuzco:
## Changes
- attempt to load token from local auth provider first before kiwix provider. This is because the local auth provider loads it's token from local storage and we can immediately detect what was the last auth mode as opposed to trying ouath provider first before local.
- store page where user was before they click on sign-in so that we can redirect them back to there after they successfully sign in
- invalidate any in-flight calls to refresh token during logout
- optionally initialize auth provider based on configured providers

This closes #414

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -184,7 +184,7 @@ const router = createRouter({
   routes,
 })
 
-// Add global navigation guard to update title
+// Add global navigation guard to update title and handle auth redirects
 router.beforeEach((to, from, next) => {
   // Update document title
   if (to.meta.title) {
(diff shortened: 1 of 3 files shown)


benoit74 (commented):
I agree this solves an important problem and LGTM.

However I don't get how this fully solves #414: when ZF API is unavailable, don't we still have "bad" errors which prevent UI from rendering properly?

elfkuzco:
> However I don't get how this fully solves https://github.com/openzim/cms/issues/414: when ZF API is unavailable, don't we still have "bad" errors which prevent UI from rendering properly?

See https://github.com/openzim/cms/issues/414#issuecomment-4912874725

benoit74 (approved):
Indeed, LGTM

PR #422, opened 2026-07-13
encapsulate location kinds in status filter

### LINKED ISSUE #420: Too much information about deleted books is displayed
Image

Looking at the above ZIM file I see two files that should have been gone for a few days already:
- if they're still displayed then they haven't been deleted;
- if the have not been deleted why would they be marked as such?

My view here is that only books that can be acted up (=moved or deleted) should be displayed, otherwise it is simply confusing.

elfkuzco:
## Rationale
The API already allows to filter books by multiple location kinds but UI only allows user to select only one location. Also, not all locations are available to the user primarily because `to_delete` and `to_delete` aren't "locations". This PR gives the filter a better name "Status" and adds options that have better meaning

Screenshot_20260713_010728
Screenshot_20260713_010714
Screenshot_20260713_010120

This closes #420

### DIFF
--- a/frontend/src/components/BookTable.vue
+++ b/frontend/src/components/BookTable.vue
@@ -166,7 +166,7 @@ interface Props {
   loadingText?: string
   filters?: {
     name: string
-    location_kind: string
+    status: string
     flag: string
   }
   isServerSide?: boolean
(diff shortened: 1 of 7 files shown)


benoit74 (approved):
LGTM

PR #424, opened 2026-07-14
show status of title modification events in UI

### LINKED ISSUE #419: Better indicate to user when title is being modified and when it completed
From https://github.com/openzim/cms/issues/417

We should display something more clearly to the user when a background event was queued for his action. This is the case for background file move operations but not for title modifications.

When user modifies title name, it is very hard to guess that an async event has been queued, and complex to check it did complete

I think we should return info about the fact that an event has been queued, display a spinning wheel + text indicating background event has been fired, and then display when it completed, and only then refresh UI. 

The "events" tab in inbox is a useful thing for admins, but it does not sufficiently help users to understand what is going on when title name is modified.

elfkuzco:
## Changes
- filter events by the ID of the title
- show status of background event to process title in UI

Screenshot_20260714_110944

This closes #419

### DIFF
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -11,4 +11,6 @@ export default {
   NOTIFICATION_DEFAULT_DURATION: 5000, // 5 seconds
   NOTIFICATION_ERROR_DURATION: 8000, // 8 seconds for errors
   NOTIFICATION_SUCCESS_DURATION: 3000, // 3 seconds for success
+  // Background event polling
+  BACKGROUND_EVENT_POLL_INTERVAL: 2000, // 2 seconds
 }
--- a/backend/src/cms_backend/api/routes/events.py
+++ b/backend/src/cms_backend/api/routes/events.py
@@ -1,4 +1,5 @@
 from typing import Annotated
+from uuid import UUID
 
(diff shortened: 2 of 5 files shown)


frontend/src/constants.ts
```diff
@@ -11,4 +11,6 @@ export default {
   NOTIFICATION_DEFAULT_DURATION: 5000, // 5 seconds
   NOTIFICATION_ERROR_DURATION: 8000, // 8 seconds for errors
   NOTIFICATION_SUCCESS_DURATION: 3000, // 3 seconds for success
+  // Background event polling
+  BACKGROUND_EVENT_POLL_INTERVAL: 60000, // 60 seconds
```
benoit74: Let's be a bit more aggressive with 2 seconds interval. This is quite rare to have a `title_modified` event and the quicker we are to report update status the less frustrated user will be

benoit74 (changes requested):
Since it takes up to 1 minute (by default) to solve `title_modified` events, I'm quite concerned user will keep refreshing the page to get updated status while it is not needed.

Shall we display something in the UI indicating everytime we are refreshing the status? Or if too complex, simply change wording to indicate status will update automatically.

benoit74 (approved):
LGTM

PR #425, opened 2026-07-16
load offliners last on component mount

### LINKED ISSUE #414: CMS UI interactions with Zimfarm should be polished

elfkuzco:
## Changes
- load offliners list from zimfarm last after components mount

This fixes #414

### DIFF
--- a/frontend/src/components/BooksBaseView.vue
+++ b/frontend/src/components/BooksBaseView.vue
@@ -312,7 +312,6 @@ watch(
 )
 
 onMounted(async () => {
-  await offlinerStore.fetchOffliners()
   loadingFlavours.value = true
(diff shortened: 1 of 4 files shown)


benoit74 (approved):
LGTM

PR #434, opened 2026-07-16
promote book with empty actions if possible

### LINKED ISSUE #429: Directly move book to production when promotion has no action needed
When book promotion dry run returns no action to perform, rather than only displaying a "Cancel" button, we should also still show a "Promote" button to directly move book to production, since moving book to production is anyway part of the promotion when there are actions needed.

Image

elfkuzco:
## Changes
- remove restriction to reject promotion when there are no actions to perform
- show button to promote book in UI when there are no actions to perform

This closes #429


benoit74 (approved):
LGTM

PR #435, opened 2026-07-16
remove alert button to edit book flavour

### LINKED ISSUE #432: Remove box inviting to edit book or title when there is a flavour mismatch
The orange box below is not useful anymore at all:

Image

It used to be a nice shortcut to edit book flavour or title flavours. 

Editing title flavours is not possible anymore now that we have the "Promote" button.

Editing book flavour is a very niche case we should not put so much forward (in general, book flavour should not be updated).

I hence propose to remove the whole "orange box".

elfkuzco:
## Changes
- remove alert button to edit book flavour

This fixes #432

### DIFF
--- a/frontend/src/views/BookView.vue
+++ b/frontend/src/views/BookView.vue
@@ -99,33 +99,6 @@
         </div>
       </v-alert>
 
-      <v-alert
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #436, opened 2026-07-20
expose API to unpromote prod book to staging

### LINKED ISSUE #433: Remove "Move book" button and API
When moving a book from staging to production, we should not use the "Move book" button anymore but rather the "Promote book" button, so that what needs to be fixed is fixed, and what needs to be confirmed is confirmed. The "Move button" should be hidden for this case.

When moving a book from production to staging, I would prefer we have a similar "Unpromote book" button, so that we have (a-)symmetric buttons.

We should also remove `/books/{book_id}/move` API and add a `/books/{book_id}/unpromote` API.

We finally need to ensure that `/books/{book_id}/promote` works only when book is in quarantine or staging, and `/books/{book_id}/unpromote` works only when book is in production.

elfkuzco:
## Changes
- remove endpoint to move book from prod/staging to prod/staging
- add API to unpromote prod book to staging
- refactor UI to change calls to move book to use promote/unpromote book actions

Screenshot_20260720_101311

This closes #433

### DIFF
--- a/backend/src/cms_backend/api/routes/books.py
+++ b/backend/src/cms_backend/api/routes/books.py
@@ -1,5 +1,5 @@
 from http import HTTPStatus
-from typing import Annotated, Literal
+from typing import Annotated
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, Path, Query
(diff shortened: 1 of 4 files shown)


benoit74 (approved):
LGTM

PR #437, opened 2026-07-20
mark title flavours as rotten

### LINKED ISSUE #397: Detect and display rotten titles
We need some solution to see rotten titles, i.e. titles for which a given title flavour has not been updated for very too long. Problem could be a title or flavour level obviously.

For that, we need to:
- add a filter in title view to search for rotten titles (titles which have a flavour which has not seen a new book for too long)
- just like article and media count, allow to customize the threshold to consider a flavour as "too old" per collection (with a global default in environment variable and if not set 13 months)
- enhance title flavour display (from https://github.com/openzim/cms/issues/396) to show when each flavour has been updated, and wether it is too old or not

### LINKED ISSUE #427: Enhance display of title flavours
Display of title flavours is a bit too rough, especially when we see only a single flavour. 

Image

I see two alternatives:
- use cards instead of a "table" list
- add "table" headers (flavour, Zimfarm recipe)

I let you choose the one you prefer to implement.

elfkuzco:
## Changes
- display title flavours as cards
- track when a book belonging to a specific flavour was added to a title
- set env to determine if a title flavour is rotten. flavour is rotten if no book with matching flavour was added
- filter books by status: rotten, active
- move individual logic to apply book actions to dedicated functions

Screenshot_20260720_125610
Screenshot_20260720_123239

This closes #397 
This fixes #427


backend/src/cms_backend/context.py
```diff
@@ -142,3 +142,9 @@ class Context:
     zimfarm_api_url: str = field(
         default=os.getenv("ZIMFARM_API_URL", "https://api.farm.openzim.org/v2")
     )
+
+    rotten_flavour_threshold: timedelta = field(
+        default=timedelta(
+            seconds=parse_timespan(os.getenv("ROTTEN_FLAVOUR_THRESHOLD", default="52w"))
```
benoit74: Default should be 13 months (if not possible to use months then use `56w`)

benoit74 (approved):
LGTM, please change default value and merge, no need for a new review pass.

PR #438, opened 2026-07-20
show buttons to set title metadata from title

### LINKED ISSUE #431: Update of title metadata during promotion is confusing and/or broken
When promoting https://cms.openzim.org/book/e846e4de-1951-b5bd-05f3-9a2592943394 to production, it first shows a list of title metadata, but it is hard to know if these are updated values (coming from the book we are promoting) or current title values (which are "outdated").

Image

Then, when we confirm the promotion we see a diff which display something more interesting.

Image

But both seem to omit the "License" metadata which should be updated as well according to https://cms.openzim.org/title/stackoverflow.com_en_all/edit

Image

We need to polish this to have a great UX and obviously update all title metadata as needed.

elfkuzco:
## Changes
- include in message that updates to title metadata are coming from book
- show difference between title metadata and book metadata and show button to use title metadata instead of book metadata
- include `License` and `LongDescription` book metadata

Screenshot_20260720_154659
Screenshot_20260720_154641

This fixes #431

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -419,7 +419,7 @@ def recover_book(session: OrmSession, book_id: UUID) -> Book:
 def get_differing_metadata_keys(book: Book) -> list[str]:
     """Get the list of metadata keys that are different between book and it's title.
 
-    Assumes book and title both have mandatory metadata set.
+    Assumes book has mandatory metadata set.
     Assumes that the book name and title name already match, thus aren't checked.
     """
 
@@ -433,6 +433,10 @@ def get_differing_metadata_keys(book: Book) -> list[str]:
         "Description": book.zim_metadata["Description"],
         "Language": book.zim_metadata["Language"],
         "Illustration_48x48@1": book.zim_metadata["Illustration_48x48@1"],
+        "LongDescription": book.zim_metadata.get("LongDescription"),
+        "License": book.zim_metadata.get("License"),
+        "Relation": book.zim_metadata.get("Relation"),
+        "Source": book.zim_metadata.get("Source"),
     }
 
     title_metadata = {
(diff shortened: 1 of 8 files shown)


benoit74 (changes requested):
I realize this only now but we need to update all title metadata (we miss creator, publisher, source, etc ...), and we should display only metadata which are different.

What if the backend sends only metadata which are different, and both book + title metadata values, wouldn't this help? We could even send a "kind" (text or image) so that rendering in UI is simplified.

elfkuzco:
> What if the backend sends only metadata which are different, and both book + title metadata values, wouldn't this help? We could even send a "kind" (text or image) so that rendering in UI is simplified.

Not sure how the "kind" simplifies things because some text fields have constraints and some don't

elfkuzco:
> I realize this only now but we need to update all title metadata (we miss creator, publisher, source, etc ...), and we should display only metadata which are different.

I just realized that only the mandatory metadata fields cause a book metadata mismatch which the action is trying to solve. All these other fields aren't really consequential in solving the issue which is what the action is trying to solve. Should I still proceed?

benoit74:
> I just realized that only the mandatory metadata fields cause a book metadata mismatch which the action is trying to solve. 

OK then we should solve this as well. All metadata fields should cause a book metadata mismatch, sorry for having overlooked this so far.

> Not sure how the "kind" simplifies things because some text fields have constraints and some don't

If it does not help, then don't do it.

elfkuzco:
Did a bit of refactoring (mostly moving every action to a function) in addition to optionally returning/rendering only the different metadata.
- included a new component to show metadata differences
- included buttons to use all title values
- show (no value) when title has `null`. applies equally to title edit form when book metadata is null

Screenshot_20260721_112429

benoit74 (approved):
LGTM

PR #441, opened 2026-07-22
restrict access to collection, titles and books with collection permissions

### LINKED ISSUE #398: Add distinction between global editors and collection editors, and admins
We need a distinction between global editors and collection editors:
- global editors are allowed to interact with all collections
- collection editors need to be granted access to one or more collection

We also need an admin role which is needed for operations on zimfarm notifications + users.

We need to implement this collection editors permission management in the whole CMS, like it has impact on book and titles permissions.

A user with `collection_editor` role but not allowed on any collection has (at this stage) no more rights than a viewer. It is just a "validation" that he is allowed to have rights on one or more collections.

Assignment of this role + collections is done by admins. I think we should do that in users management screen + API, next to the role assignment.

elfkuzco:
## Changes
- add new roles `admin` and `collection-editor`. `admin` accounts have CRUD permissions on `zimfarm_notification`, `account` but only read permissions on all collections which allows them to select collections to add to a user they are updating. `collection-editor` role can perform CRUD actions on `book`, `title` and `collection`
- add new table `collection_permission` for storing the collections a user can access
- alter queries to retrieve books, titles and collection to take an optional parameter `accessible_collection_ids`. When it is set to None, user can view those resources (existing behaviour). When it is not None, records are filtered to ensure they belong to such collections. Rationale for making `None` imply global access as opposed to passing the list of all collection IDs is that we might want to view/perform actions on books that don't have a title (and as a consequence don't belong to any collection). Passing the list of all collections would prevent the user from ever acting on such books. This is particularly useful in inbox books which have filter for books that are pending title association
- add dependency at API level that retrieves the accessible_collection_ids for every API request
- update UI to allow setting updating user with new roles

This closes #398


backend/src/cms_backend/db/book.py
```diff
@@ -63,6 +74,12 @@ def get_book_or_none(
             (Book.needs_processing.is_(needs_processing)) | (needs_processing is None),
             (Book.has_error.is_(has_error)) | (has_error is None),
             (Book.location_kind.in_(locations or [])) | (locations is None),
+            Book.title_id.in_(
```
benoit74: Why retrieving all accessible collection titles IDs rather than just joining on Title + CollectionTitle and checking `CollectionTitle.collection_id.in_(accessible_collection_ids or [])`? The join on title and collection title is "natural" (like it is done by ID so cheap) while listing all accessible titles looks like an expensive DB operation (could return thousands of books) and a complex thing to read.

backend/src/cms_backend/db/collection.py
```diff
@@ -32,56 +35,96 @@
 
 
 def get_collection_by_id_or_none(
-    session: OrmSession, library_id: UUID
+    session: OrmSession,
+    library_id: UUID,
```
benoit74: Please rename from `library_id` to `collection_id`, we've overlooked this imprecision in the past. (here and in other collection-related methods)

backend/src/cms_backend/db/collection.py
```diff
@@ -281,9 +340,13 @@ def create_collection(
             ) from exc
         logger.exception("Unknown exception encountered while creating collection")
         raise
+
+    create_collection_permission(session, collection.id, author_id)
```
benoit74: See other change on roles, but I don't think we need to create a collection permission for author since creating a collection will be possible only for admins + editors.

backend/src/cms_backend/db/collection_permission.py
```diff
@@ -0,0 +1,47 @@
...
+    """Get the collection IDs account is allowed to operate on
+
+    NOTE: None implies to skip check if account has permission to collection.
+    This translates to account having access to all collections, books, titles, etc
+    """
+    if account is None:
+        return []
+    if account.role in [RoleEnum.EDITOR, RoleEnum.ADMIN]:
+        return None
+    return session.scalars(
```
benoit74: Collection permissions should be allowed only on `COLLECTION_EDITOR` role for now, I do not want collection permissions added inadvertently on another role to be able to have any effect.

backend/src/cms_backend/db/flavour.py
```diff
@@ -6,14 +7,31 @@
...
+    skip: int = 0,
+    accessible_collection_ids: Sequence[UUID] | None = None,
 ) -> ListResult[TitleFlavourSchema]:
-    stmt = select(TitleFlavour).where(TitleFlavour.title_id == title_id)
+    stmt = (
+        select(TitleFlavour)
+        .join(Title, TitleFlavour.title_id == Title.id)
+        .where(
+            TitleFlavour.title_id == title_id,
+            Title.id.in_(
```
benoit74: Same question, why not directly checking `CollectionTitle.collection_id` is in accessible collections?

backend/src/cms_backend/db/title.py
```diff
@@ -130,62 +131,137 @@ def create_title_light_schema(title: Title) -> TitleLightSchema:
...
+    accessible_collection_ids: Sequence[UUID] | None = None,
+) -> Title:
     """Get a title or raise RecordDoesNotExistError if it doesn't exist."""
-    if (title := get_title_by_name_or_none(session, name=name)) is None:
+    if (
+        title := get_title_by_name_or_none(
+            session, name=name, accessible_collection_ids=accessible_collection_ids
+        )
+    ) is None:
         raise RecordDoesNotExistError(f"Title with name '{name}' does not exist")
```
benoit74: We should change all these errors to say something more accurate like "does not exist or is not accessible to you"

backend/src/cms_backend/schemas/models.py
```diff
@@ -72,16 +72,31 @@ class FileLocation:
...
+class BaseAccountCreateUpdateSchema(BaseModel):
+    username: NotEmptyString | None = Field(default=None, min_length=3)
+    display_name: NotEmptyString | None = Field(default=None, min_length=3)
+    collections: list[NotEmptyString] | None = None
+    role: RoleEnum | None = None
+    idp_sub: UUID | None = None
+
+    @model_validator(mode="after")
+    def restrict_collections_to_collection_editor(self) -> Self:
+        if (self.role == RoleEnum.COLLECTION_EDITOR and self.collections is None) or (
```
benoit74: Please split the "if" in two to have clearer errors:
- "Collections must be specified when setting role to COLLECTION_EDITOR"
- "Collections must not be specified when role is not COLLECTION_EDITOR"

And allow collections to be an empty list when role is not COLLECTION_EDITOR (None or empty list are both fine)

backend/src/cms_backend/roles.py
```diff
@@ -24,19 +24,31 @@ class TitlePermissions(ResourcePermissions):
 
 class RoleEnum(StrEnum):
     EDITOR = "editor"
+    COLLECTION_EDITOR = "collection-editor"
     VIEWER = "viewer"
     ZIMFARM = "zimfarm"
+    ADMIN = "admin"
 
 
 ROLES: dict[str, dict[str, dict[str, bool]]] = {
```
benoit74, line 33: Sorry, looks like I've been a bit imprecise here:
- we want ADMIN to have more permissions than EDITOR, and be the only ones allowed to interact with accounts
- we want EDITOR + COLLECTION_EDITOR to not be allowed to create or delete collections, only ADMIN should be allowed (this is something I've overlooked)
- EDITOR should still have "read" permission on accounts to be able to list them and check their permissions, this is OK

I.e. something like that if I'm not mistaken:

```
ROLES: dict[str, dict[str, dict[str, bool]]] = {
    RoleEnum.VIEWER: {},
    RoleEnum.ADMIN: {
        "book": ResourcePermissions.get_all(),
        "title": TitlePermissions.get_all(),
        "zimfarm_notification": ResourcePermissions.get_all(),
        "account": ResourcePermissions.get_all(),
        "collection": ResourcePermissions.get_all(),
    },
    RoleEnum.EDITOR: {
        "book": ResourcePermissions.get_all(),
        "title": TitlePermissions.get_all(),
        "collection": ResourcePermissions.get_all(read=True, update=True),
        "account": ResourcePermissions.get_all(read=True),
    },
    RoleEnum.COLLECTION_EDITOR: {
        "book": ResourcePermissions.get_all(),
        "title": TitlePermissions.get_all(),
        "collection": ResourcePermissions.get_all(read=True, update=True),
    },
    RoleEnum.ZIMFARM: {
        "zimfarm_notification": ResourcePermissions.get(read=True, create=True),
    },
}
```

frontend/src/constants.ts
```diff
@@ -21,7 +21,7 @@ export default {
   COOKIE_LIFETIME_EXPIRY: '10y', // 10 years
   TOKEN_STORAGE_KEY: 'cms-auth',
   // User roles
-  ROLES: ['editor', 'viewer'] as const,
+  ROLES: ['editor', 'viewer', 'collection-editor', 'admin'] as const,
```
benoit74: Change order to `viewer`, `collection_editor`, `editor`, `admin` to sort them by "power"

benoit74 (changes requested):
See inline comments + please rename "editor" role to "global_editor" for better clarity.

backend/src/cms_backend/db/book.py
elfkuzco: I just tried it this way but I realize it causes the queries that `get_one_or_none()` to fail with `sqlalchemy.exc.MultipleResultsFound: Multiple rows were found when one or none was required`. This happened in a couple of test cases because a test case associates a title with multiple collection titles. While this happened for a case to fetch title, it can apply to books too belonging to such titles. Solution is to use `distinct()` across such places to deduplicate

elfkuzco: Probably unlikely to have a title belonging to multiple collections in prod but it's worth mentioning

EDIT: Wrong assumption. Found a couple in prod DB dump

benoit74: We need anyway to have code ready for a title belonging to multiple collections. It is an edge case but it must work correctly. We can probably write the query differently, the "Join" operation is probably not the appropriate query, we can maybe solve this with `exists` criteria or another solution.

benoit74 (approved):
LGTM, thank you !

PR #442, opened 2026-07-22
allow collections to be private/public

### LINKED ISSUE #399: Add distinction between public and private collections
We need a distinction between public and private collections.

This is a setting at collection level.

Public collection(s) are typically Kiwix, Private collection(s) are typically BSF, Endless, ...

Private collections are not something secret or very strictly controlled like a bank account details, they are only here to not show in the public some work we consider should not be public. A bit like a fence in your garden or a curtain at home. Not that it must be very strictly controlled, but we prefer people do not see what is behind the curtain, typically because orgs have paid for this work and do not want this to be publicly exposed in uncontrolled manners.

Collection, Title and Book APIs inherit permissions from the collection.

Private collections are not visible to users without collection_editor or above right (see https://github.com/openzim/cms/issues/398), including public APIs (e.g. `GET /books`).

For title and collection endpoints, it is not possible to use the name as identifier when collection is private, only ID is allowed.

Consequence on staging APIs still has to be clarified (no change needed in this issue).

elfkuzco:
## Changes
- add new property to collections `is_private` for controlling visibility status
- send is_private when creating/updating collection
- allow collections to be filtered by visibility status
- allow unauthenticated users, viewers and zimfarm roles to access public collections. they are still constrained by the permissions that come with their role to perform CRUD operations on resources

Screenshot_20260722_154806
Screenshot_20260722_154750

This closes https://github.com/openzim/cms/issues/399


benoit74 (changes requested):
I think this PR misses adaptations for two statements of the issue.

> For title and collection endpoints, it is not possible to use the name as identifier when collection is private, only ID is allowed.

I don't see any change for that.

> Operations for which we do not know the collection yet (e.g. creating a new title) obviously mandate to have global_editor or above permission

Let's say "create_title" permission is available only to GLOBAL_EDITOR and ADMIN role, this is probably fine at least in a first approximation.

elfkuzco:
For the first point, I have ensured all calls from route level use the high-level functions `get_(collection|tilte)_or_none` and `get(collection|title)`. The ones that access by name/ID are still useful for mill, mill and maint-scripts. Also updated the UI to make all calls with UUID as it avoids the overhead of checking if collection is private or not.

Regarding this:
>  Operations for which we do not know the collection yet (e.g. creating a new title) obviously mandate to have global_editor or above permission
> Let's say "create_title" permission is available only to GLOBAL_EDITOR and ADMIN role, this is probably fine at least in a first approximation.

Admin already has all permissions and global editor (at least based on https://github.com/openzim/cms/pull/441#discussion_r3636577145) already has all permissions on title.

benoit74:
> Admin already has all permissions and global editor (at least based on https://github.com/openzim/cms/pull/441#discussion_r3636577145) already has all permissions on title.

My point was that collection_editor should not have permission to create titles.

elfkuzco:
I have updated the code to give them `read` and `update` permissions since they will want to promote book. Even though they don't have permissions to create title, book promotion  action triggered by them could create title. For now, the action requires `update` permission on books and titles

benoit74 (approved):
LGTM

PR #445, opened 2026-07-24
use existing books filename property to compute new filenames

### LINKED ISSUE #444: Computation of book filename should also take into account deleted books
Looks like computation of book filename does not take into account deleted books. This is not ideal since it might create misunderstanding (by reusing same filename with different content).

I suspect this is linked to the fact that we use book location instead of the book itself (filename is stored in book table only quite recently).

elfkuzco:
## Changes
- use existing books filename property to compute new filenames

This fixes #444

### DIFF
--- a/backend/src/cms_backend/utils/filename.py
+++ b/backend/src/cms_backend/utils/filename.py
@@ -2,12 +2,13 @@
 
 import re
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM

PR #447, opened 2026-07-24
only generate diff of action data keys in update_title_metadata action

### LINKED ISSUE #446: Promote book confirmation UI has misleading content
When I'm about to promote a book which has only illustration change (for instance, I suspect the issue is wider) then the confirmation popup displays worrying changes of title metadata to `null`. These changes do not apply when we confirm, so the fact they are displayed is both wrong and misleading the user.

See example below:

Image

elfkuzco:
## Changes
- only generate diff of the different metadata keys in update_title_metadata action as backend now returns the different keys and not the entire metadata

This fixes #446

### DIFF
--- a/frontend/src/components/PromoteBookDialog.vue
+++ b/frontend/src/components/PromoteBookDialog.vue
@@ -567,29 +567,11 @@ const titleDiffs = computed(() => {
 
     switch (action.kind) {
       case 'update_title_metadata': {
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM, I like a lot the fact it is more concise and extensible (no need to modify code anytime we add a new metadata)

PR #448, opened 2026-07-28
update recipe flags on zimfarm

### LINKED ISSUE #333: Better solve book metadata issue

elfkuzco:
## Rationale
This PR adds functionality to the UI to update flags responsible for recipe metadata from the CMS. During book promotion actions, if there are actions to update title metadata, UI loads recipe details and informs user to make corresponding changes to the zimfarm recipe that created the book

## Changes
- prompt user to update appropriate flags that are responsible for the metadata on zimfarm
- add methods to authenticate with zimfarm in cms auth store
- upload base64 blobs that are responsible for `Illustration` metadata to zimfarm before applying the patch with the returned URL

Screenshot_20260728_031111
Screenshot_20260728_031119
Screenshot_20260728_031138
Screenshot_20260728_031148

This closes #333

### DIFF
--- a/frontend/src/components/RecipeUpdateDialog.vue
+++ b/frontend/src/components/RecipeUpdateDialog.vue
@@ -1,0 +1,483 @@
+<template>
+  <v-dialog v-if="canShow" v-model="isOpen" max-width="700px" persistent>
+    <v-card>
+      <v-card-title class="text-h5 pa-4 bg-primary">
+        <span class="text-white">Update Recipe on Zimfarm</span>
+      </v-card-title>
+
+      <v-card-text class="pt-4">
+        <v-alert
+          v-if="error && !permissionDenied"
+          type="error"
+          variant="tonal"
+          class="mb-4"
+          closable
+          @click:close="clearError"
+        >
+          {{ error }}
+        </v-alert>
+
+        <div v-if="checkingAuth || loading" class="text-center pa-8">
+          <v-progress-circular indeterminate size="48" />
+          <div class="mt-3 text-body-2 text-medium-emphasis">
+            {{ loadingMessage }}
+          </div>
+        </div>
+
+        <div v-else-if="needsAuth">
+          <p class="text-body-2 mb-3">
+            You need to authenticate with the Zimfarm API to update the recipe.
+          </p>
+          <v-text-field
+            v-model="username"
+            label="Zimfarm Username"
+            variant="outlined"
+            density="compact"
+            class="mb-2"
+            :disabled="authLoading"
+          />
+          <v-text-field
+            v-model="password"
+            label="Zimfarm Password"
+            type="password"
+            variant="outlined"
+            density="compact"
+            class="mb-3"
+            :disabled="authLoading"
+            @keyup.enter="authenticate"
+          />
+        </div>
+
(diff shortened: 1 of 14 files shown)


frontend/src/components/RecipeUpdateDialog.vue
```diff
@@ -0,0 +1,478 @@
...
+  const blobDataKeys = new Set<string>()
+
+  for (const zimMeta of props.defSpec.schema.zimMetadata) {
+    const cmsField = ZIM_METADATA_TO_CMS_FIELD[zimMeta.metadata]
+    if (!cmsField || newData[cmsField] === undefined) continue
+
+    const flag = props.defFlag.flags.find((f) => f.key === zimMeta.flag)
+    if (!flag) continue
+
+    oldData[flag.data_key] = props.recipeMetadata[cmsField] ?? null
```
benoit74, line 213: When recipe flag is not set, then we should ignore this flag as well (like not consider it as a difference) because it means the book metadata has been automatically computed by the scraper, and this is very probably what the Zimfarm editor wants.

frontend/src/components/RecipeUpdateDialog.vue
```diff
@@ -0,0 +1,478 @@
...
+        </v-btn>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+
+  <!-- Confirmation dialog shown before asking for credentials -->
+  <ConfirmDialog
+    v-model="showAuthConfirm"
+    title="Update Recipe on Zimfarm"
+    confirm-text="Authenticate with Zimfarm"
```
benoit74: I would rather use a simple `Proceed` text even if in turns it will start the authentication workflow. Explaining user will have to authenticate on Zimfarm at this moment is just an unneeded detail (and might not be needed if user is already authenticated).

benoit74 (changes requested):
Thank you, looks a lot like what I expected, with even some improvements over few things I wrongly anticipated (like we always have at most one recipe to update, never two, since we are promoting a book so there is one single recipe attached ...)

frontend/src/components/RecipeUpdateDialog.vue
benoit74: @elfkuzco did you missed this comment or did I missed the code which already ensures it works like this?

elfkuzco: The last recent  commit skips including the recipe flags that are null from the diffs that are computed. It is the diffs that are used to build up the fields to select. If that diff isn't included, then, the field won't show up for selection

benoit74: Indeed, thank you

benoit74 (approved):
LGTM

PR #458, opened 2026-07-30
show all informational issues as actions

### LINKED ISSUE #450: Impossible to promote a book which has only "Informational" issues
See https://cms.openzim.org/book/6b750d2a-512c-b69f-3705-83526b73cc94

It is impossible to click the "Promote" button, probably because book has only informational issues.

### LINKED ISSUE #451: We should display all "informational" issues (or none)
See https://cms.openzim.org/book/6b750d2a-512c-b69f-3705-83526b73cc94/issues

Book has two issues but promotion dialog show only one "informational" action.

Image

Unless I forgot why we did it like this (which is possible ^^), I feel like we should be more coherent and display same amount of information on both sides.

I feel like displaying all the informational issues in the promotion popup is the correct solution.

@elfkuzco did I forgot something we've decided or is it ok to implement this change?

elfkuzco:
## Changes
- display media/article count issues as informational actions
- extend zimcheck_issue action to return message when no zimcheck URL is found or failed to compute zimcheck summary
- allow user to promote book when there are only informational actions

Screenshot_20260730_114239
Screenshot_20260730_114309

This closes #451 
This fixes #450

### DIFF
--- a/backend/src/cms_backend/db/book.py
+++ b/backend/src/cms_backend/db/book.py
@@ -797,6 +797,26 @@ def get_book_unsupported_languages(book: Book) -> list[str]:
     return unknown_languages
 
 
+def get_latest_prod_book(session: OrmSession, book: Book):
+    """Get the latest prod book that has the same title and flavour as book.
+
+    If none is found, the book is returned.
+    """
+    latest_book = session.scalars(
+        select(Book)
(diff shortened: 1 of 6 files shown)


benoit74 (approved):
LGTM

PR #464, opened 2026-07-31
reuse add_book_to_title to promote books without title

### LINKED ISSUE #463: Promoting a book from quarantine did not correctly computed the correct filename

elfkuzco:
## Changes
- move logic to `add_book_to_title` to `db/book` module and reuse when we are applying action to create title. This action is never attached with other actions that restore title or update title metadata because title never exists at that point. This way, when we are adding book to title, book filename is computed from filename along with other helpful events that get logged when book is newly added to title

This fixes #463


benoit74 (approved):
LGTM

PR #466, opened 2026-07-31
add maint-script to move book to new collection path

### LINKED ISSUE #463: Promoting a book from quarantine did not correctly computed the correct filename

elfkuzco:
## Rationale
Script proposed https://github.com/openzim/cms/issues/463#issuecomment-5143766119 does not achieve to rename book because shuttle identifies the locations as the same (same warehouse and path). This PR introduces maint-script to move book from one collection path to another.

This fixes #463

### DIFF
--- a/backend/maint-scripts/move_book_to_new_collection_path.py
+++ b/backend/maint-scripts/move_book_to_new_collection_path.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+"""Maintenance script to move prod book to new collection path in the same title.
+
+NOTE:
+- Books will be removed from existing collection paths and scheduled to be moved
+to new collection path
+- All
+"""
+
+import argparse
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


elfkuzco:
@benoit74 ,  because `CollectionTitle` have primary key made of `collection_id` and `title_id`, prod books belonging to such titles will need to be moved to new path. For the two books mentioned in #463 , their titles only have them as prod books, so this will not affect anything else.

Script can be run as (setting correct paths instead of <new-path>). Paths must exist on FS
```sh
move_book_to_new_collection_path.py --book=6c38dade-f569-fb59-4116-348b477c4f50 --collection=Kiwix --path<new-path> --generate-new-filename

move_book_to_new_collection_path.py --book=83fe4294-6296-9e74-daa7-99be7d2511f3 --collection=Kiwix --path<new-path> --generate-new-filename
```

PR #468, opened 2026-08-03
consider book filename when detecting identical target

### LINKED ISSUE #463: Promoting a book from quarantine did not correctly computed the correct filename

elfkuzco:
## Changes
- consider book filename when trying to detect if book current location and target location are identical. This allows a book to be renamed within the same path. As a consequence, defer updating the book status since "status" is part of the composite primary key. Only when we have copied all files (or deleted identical paths) do we set all the locations as current

This fixes #463


benoit74 (approved):
LGTM

PR #469, opened 2026-08-04
upload zim to collection and create zimfarm task

### LINKED ISSUE #322: Allow to upload manually a ZIM

elfkuzco:
## Rationale
This PR enhances the CMS by allowing editors to upload an already created zim from their local filesystem. The CMS on receiving such files creates the task on zimfarm which will advertise the book back to the CMS once checks and processing are done by a worker

## Changes
- allow users to upload zims manually. uploads are done as S3 multipart uploads and API generates presigend URLs for uploading and executing the task. Presigned URLs for uploading expire after one day whereas the ones used by the worker expire after 7 days.
- add logic to authenticate with zimfarm API
- add task_id to book and corresponding UI link to view task on Zimfarm
- show tasks in collections views (where they were uploaded from)
- add shuttle task to delete zims of completed tasks
- add mill task to update status of requested tasks

Screenshot_20260804_114244
Screenshot_20260804_114257

This closes #322


elfkuzco:
@benoit74 , if PR is approved, we will need to set policies on Wasabi to delete aborted/incomplete uploads and possibly retention policy on bucket too.

elfkuzco:
Because uploads are done by editors, I have opted to keep the upload section in the collections page so collection_editor and global_editor can upload zims. But i'm stuck on the issue when book arrives and there's no title for it in which case `collection_editor` will not be able to act on it. Even worrying is that global editor when they try to promote book might attach book to a collection that is different from what the uploader intended. How do we go about this? Do we want to enhance promote book logic to only allow edits to the collection title when none is set? This way, when we determine book came from an uploaded zim, then, we fix it to that collection and path

@benoit74 , wdyt?

benoit74:
> Because uploads are done by editors, I have opted to keep the upload section in the collections page so collection_editor and global_editor can upload zims. But i'm stuck on the issue when book arrives and there's no title for it in which case collection_editor will not be able to act on it. Even worrying is that global editor when they try to promote book might attach book to a collection that is different from what the uploader intended. How do we go about this? Do we want to enhance promote book logic to only allow edits to the collection title when none is set? This way, when we determine book came from an uploaded zim, then, we fix it to that collection and path

Good point ; can you develop what you mean by "enhance promote book logic to only allow edits to the collection title when none is set"?

elfkuzco:
Screenshot_20260804_170119
Screenshot_20260804_170109


Here's the UI when API sets collection titles in create_title action

elfkuzco:
I realized we need the recipe_id in order to be able to reconstruct this path. So, I think maybe we shouldn't delete recipe till zimfarm has published book to CMS or mill has detected task has failed/canceled.

kelson42:
> I realized we need the recipe_id in order to be able to reconstruct this path. So, I think maybe we shouldn't delete recipe till zimfarm has published book to CMS or mill has detected task has failed/canceled.

sorry to intervene here as @benoit74 is out for a few days, but here there onviously no recipe-id to link to the book/title.

elfkuzco:
> sorry to intervene here as @benoit74 is out for a few days, but here there onviously no recipe-id to link to the book/title.

Zimfarm advertises the recipe_id as part of the payload that gets sent in the CMS notification. And before we create the task on zimfarm, we create a recipe and store the recipe_id. But, I'm proposing to delay the deletion of the recipe on zimfarm (which is currently done immediately after creation) till it gets advertised to CMS.

benoit74 (changes requested):
I did not had time to go through all details, but I have some preliminary feedback.

First, I'm now convinced we should have a zimfarm in the CMS docker-compose, just like we have one in WP1 ; probably mostly all components to test from end-to-end ; it is just too tedious to test the integration without it, too many bits to move to make it works ; please add it to the PR and ensure docker-compose default env values are correct to target this local CMS-tied Zimfarm.

Second, I feel like there is a conception issue. Reading again the issue notes, I realize that my last idea was to have the "Upload" function in the "TitleDetail" not in "CollectionDetail". Uploading at collection level adds just too much complexity. Not sure why you implemented it at that level. The problem is that we can be pretty sure that uploaded ZIM name has a significant chance to not match the title name we wanna use. We hence cannot say that the ZIM will follow the standard process, and we need to tie the upload to the target title. The user hence upload a ZIM to a given title and we "force" it to match. We should still be complaining when ZIM names does not match title name, but the book is automatically attached to proper title when it arrives.

WDYT? Do I miss something obvious we've already discussed?

Finally, regarding the fact that recipe will have been deleted when book arrives in the CMS, let's change our strategy and follow WP1 path: I realize we should never delete the Zimfarm recipe and reuse it over and over (creating it on first upload, and then only changing the URL each time a ZIM is uploaded). This gives the same benefits as WP1: full title history in Zimfarm in a single recipe. And also in this case better and simpler link between Zimfarm and CMS.

elfkuzco:
> Second, I feel like there is a conception issue. Reading again the issue notes, I realize that my last idea was to have the "Upload" function in the "TitleDetail" not in "CollectionDetail". Uploading at collection level adds just too much complexity. Not sure why you implemented it at that level. The problem is that we can be pretty sure that uploaded ZIM name has a significant chance to not match the title name we wanna use. We hence cannot say that the ZIM will follow the standard process, and we need to tie the upload to the target title. The user hence upload a ZIM to a given title and we "force" it to match. We should still be complaining when ZIM names does not match title name, but the book is automatically attached to proper title when it arrives.

> WDYT? Do I miss something obvious we've already discussed?

Rationale for keeping it in the collection detail view is that collection editors do not have permission to create a title. So, if we keep it in the title detail view, this would be based on the assumption the title already exists. But that would mean, collection editor cannot upload zim because they cannot create title. Also, I was assuming a title can belong to many collections (via multiple collection title). Which collection permissions do we want to use assuming one is private and one is public? If we take the public route and zim arrives, then all collection editors can see the book that is pending association to a title.

benoit74:
> Rationale for keeping it in the collection detail view is that collection editors do not have permission to create a title. So, if we keep it in the title detail view, this would be based on the assumption the title already exists. But that would mean, collection editor cannot upload zim because they cannot create title. 
Also, I was assuming a title can belong to many collections (via multiple collection title). Which collection permissions do we want to use assuming one is private and one is public? If we take the public route and zim arrives, then all collection editors can see the book that is pending association to a title.

Given current usage, I feel like it would make more sense to have global editors create the title, assign it to proper collection and path (because they are the ones in the know of these details, especially the fact that title name being unique across all collections, choosing it wisely and ensuring it matches our convention is very important), and then tell the person who wants to upload a ZIM manually "hey, I have this title ready for you to upload ZIMs to, please proceed". Also the fact that collection editor cannot create a title should not be "bypassed" by the ability to upload a ZIM manually which would create a title automatically for them. And this limitation is not "written in stone" IMHO, this might change in the future.

Regarding titles associated to both a public and a private collection, we know the title and its books are in fact public, because public wins over private ATM.

I don't get the point about book pending association to a title. A book pending association to a title is normally hidden to everybody but global-editors and admins. And with what I propose, ZIMs manually uploaded must anyway have a strong link to the title they have been uploaded to, because I'm quite sure we will have bad ZIM names in books arriving, so we should not trust the usual ZIM name <=> title name association, it will break too often.

elfkuzco:
> Finally, regarding the fact that recipe will have been deleted when book arrives in the CMS, let's change our strategy and follow WP1 path: I realize we should never delete the Zimfarm recipe and reuse it over and over (creating it on first upload, and then only changing the URL each time a ZIM is uploaded). This gives the same benefits as WP1: full title history in Zimfarm in a single recipe. And also in this case better and simpler link between Zimfarm and CMS.

I noticed we couldn't request another task while current one was being requested. Given the uploads are now in the title view, I have constructed the recipe names as `zimwright_{title.name}` (maybe I should have used title ID instead). This way, users can upload to other title details. 

In logic to promote book dialog, should we use the name from zim metadata or the one from which the zim was uploaded?

elfkuzco:
I also realized that during the integration of zimfarm components in cms, worker running and zimfarm advertising back to cms would cause shuttle to fail because both don't share the same local paths. i was thinking of worker uploading zim to s3 and shuttle proceeding as usual. but our current shuttle implementation works only with local files. do we want to extend shuttle codebase to work with s3 too?

benoit74:
> I also realized that during the integration of zimfarm components in cms, worker running and zimfarm advertising back to cms would cause shuttle to fail because both don't share the same local paths.

They must share the same local path / volume ; or more exactly, the folder where zimfarm receiver places received files must be seen by CMS shuttle as the quarantine folder. This is how things work in production. This has to be replicated in docker-compose obviously.

> i was thinking of worker uploading zim to s3 and shuttle proceeding as usual. but our current shuttle implementation works only with local files. do we want to extend shuttle codebase to work with s3 too?

We might need to do that at some point, but I don't see a reason for not been able to replicate production setup in dev docker-compose ... and I want dev docker-compose to replicate production setup, the less different from production it is, the more bugs we will catch (or the less unrelated bugs we will experience while testing).

> I noticed we couldn't request another task while current one was being requested. Given the uploads are now in the title view, I have constructed the recipe names as zimwright_{title.name} (maybe I should have used title ID instead). This way, users can upload to other title details.

We indeed need to have one recipe per title. 

> In logic to promote book dialog, should we use the name from zim metadata or the one from which the zim was uploaded?

When arriving in the CMS, the book must automatically be associated with the title the ZIM was uploaded to. So in DB book.title_id will be set. Which means `mill.processors.book.get_matching_title()` will have to be modified to automatically match the book to proper title. Once book is matched to proper title, does CMS really still check the name? Which action is raised in promote book dialog and is it a mandatory or an optional one? 

When ZIM name does not match title name (but book has been associated to a title because it has been uploaded to it), I feel like the CMS should raise a "Metadata Mismatch", and then an optional action to update title name.

elfkuzco:
I have updated the compose file to integrate receiver and bound the receiver volumes to the shuttle path. complete interaction between cms and zimfarm now works flawlessly.

elfkuzco:
> . Once book is matched to proper title, does CMS really still check the name? 

No, CMS doesn't check it's name once book has been attached to title.

> Which action is raised in promote book dialog and is it a mandatory or an optional one?


No action is suggested if there's a mismatch between title name and book name.

benoit74:
Thank you!

Why did you kept the tasks tab in the collection details view? I feel like it would be way more useful in the title details, next to "upload". I would even suggest we have only one tab "Upload" with all information inside ; both are very closely related, and we have quite a lot of space left in both tabs.

Why did you keep asking for collection path in "Upload"? I don't think this is necessary at all.

I had a CORS problem in dev stack, looks like Zimfarm API is not allowing CMS UI (to fetch offliners among others) ; I also don't get why it allows CMS API ; probably a port mismatch

I tried to upload a ZIM and it worked pretty well up to the point where Zimfarm is trying to upload. I realized we miss the `DISABLE_WAREHOUSE_PATH: true` setting in Zimfarm API **and** in Zimfarm UI.

We can also revisit the message when upload is complete:

image

We should better indicate that upload will now be processed by the Zimfarm and CMS, and that user should watch book to arrive in title.

Requested tasks seem to never be updated, the function is not registered in `mill/main.py`. Default interval should be 1 min ; and I don't feel like building a list of requested task ids to omit makes any sense. The list will only keep growing, up to the point it will not be tractable by DB engine anymore. Why not simply passing list of statuses to omit?

CMS also failed to match the incoming book with proper title, it is stuck in quarantine "pending title".

We also need to document what needs to be setup on Wasabi S3 bucket for upload to work correctly. I'm playing with configs to confirm that aspect.

elfkuzco:
Screenshot_20260827_125610


> CMS also failed to match the incoming book with proper title, it is stuck in quarantine "pending title".

Applied this midnight but it should be working now. Also, there's been a change in the migrations, so you might want to downgrade and upgrade again. Applied all the other fixes. 

I noticed that because of this logic
https://github.com/openzim/cms/blob/beaabad218b9e0b8a31043f31b8a30372deee251/backend/src/cms_backend/db/book.py#L1341-L1346

when we associate the book to the title because it came from a requested task, it's name attempts to be updated and we can have an integrity error. Not sure how often that would come up in prod but it's worth mentioning.

benoit74:
I've opened https://github.com/openzim/cms/issues/490 to track these issues around book name vs title name ; should probably be quickly implemented after (or even before) this PR ; would this issue solve your concerns?

elfkuzco:
> I've opened https://github.com/openzim/cms/issues/490 to track these issues around book name vs title name ; should probably be quickly implemented after (or even before) this PR ; would this issue solve your concerns?

Yes

benoit74:
Thank you.

I found two issues on this screen:

Screenshot 2026-08-28 at 09 14 57

The "Status" filter is not is the right "visual block" ; should be next to the tasks, so that it is clearer that this filter allows to filter tasks and is not a mandatory thing to setup to upload a ZIM.

Immediately after upload, the text says that upload complete but there is nothing (new) shown in the list of tasks, which is rather confusing. We should immediately reload the list and requested task would appear if I'm right.

By chance, I also identified a second flaw: https://github.com/openzim/cms/issues/494 ; less keep this issue aside for now, it is not straightforward to solve at all IMHO (we do not have the ZIM ID until Zimfarm has downloaded the ZIM).

I can also confirm that https://github.com/openzim/cms/issues/490 is a pre-requisite to deploy this PR.

I also confirm that it looks like we do not need the S3 bucket to have any public permission or special CORS setting for this to work.

I tested the whole workflow and confirm it works like expected now.

elfkuzco:
> Immediately after upload, the text says that upload complete but there is nothing (new) shown in the list of tasks, which is rather confusing. We should immediately reload the list and requested task would appear if I'm right.

Okay. Tasks refresh in 1 minute intervals though

benoit74:
> Okay. Tasks refresh in 1 minute intervals though

This is OK, what I feel weird is to not have the requested task immediately appear.

elfkuzco:
added refresh after upload and used line to seperate tasks list

Screenshot_20260828_112035

backend/src/cms_backend/mill/processors/book.py
```diff
@@ -74,6 +75,22 @@ def get_matching_title(
             book.has_error = True
             return None
 
+        # if a book is created from a manually uploaded zim, we want to attach book
+        # to the title
+        if book.recipe_id:
+            requested_tasks = get_requested_tasks(
```
benoit74: Since requested tasks all have a title_id, and since the first requested task in the list will "win", I would prefer we explicitly call a `get_last_recipe_task(session, recipe_id)` and use this task `title_id` to match the book.  This would be significantly easier to read.

backend/src/cms_backend/mill/processors/book.py
```diff
@@ -74,6 +75,22 @@ def get_matching_title(
...
+                session, recipe_id=book.recipe_id
+            ).records
+            for requested_task in requested_tasks:
+                if requested_task.title_id:
+                    title = get_title_by_id(
+                        session,
+                        title_id=requested_task.title_id,
+                        accessible_collection_ids=accessible_collection_ids,
+                    )
+                    book.events.append(f"{getnow()}: found matching title {title.id}")
```
benoit74: Message should be more explicit about the fact that book has been match to title because it comes from a requested task (so that we clearly know something bad happened if this happens and was not expected).

backend/src/cms_backend/db/models.py
```diff
@@ -513,3 +517,30 @@ class Event(Base):
     created_at: Mapped[datetime]
     topic: Mapped[str]
     payload: Mapped[dict[str, Any]]
+
+
+class RequestedTask(Base):
+    __tablename__ = "requested_task"
+    id: Mapped[UUID] = mapped_column(primary_key=True)
+    recipe_id: Mapped[UUID]
+    s3_url: Mapped[str]
```
benoit74: Why storing `s3_url` in DB? I feel like this comes from Zimfarm where we had to store URL and expose it in API because it needed to be transferred to workers where we can't (or don't want to) expose the upload URL as environment variable because it is too complex to update.

Here unless I miss something, all components dealing with this URL are under our control and could use the environment variable value. And the API clients should only use presigned URL the backend will generate for them. So I would prefer we completely drop this `s3_url` to avoid any misunderstanding and strengthen security.

backend/src/cms_backend/db/models.py
```diff
@@ -513,3 +517,30 @@ class Event(Base):
     created_at: Mapped[datetime]
     topic: Mapped[str]
     payload: Mapped[dict[str, Any]]
+
+
+class RequestedTask(Base):
```
benoit74: I know it wasn't that crystal clear at this PR inception, but I feel like this would be better named `TitleUpload` ; I'm now almost certain we will always attach a ZIM upload to a title. We should update DB table, schemas and API routes... sorry for this late change ... but better now than once in production ^^

backend/src/cms_backend/shuttle/delete_uploaded_zims.py
```diff
@@ -0,0 +1,42 @@
...
+            except Exception:
+                logger.exception(
+                    "error while deleting zim upload for requested_task "
+                    f"{requested_task.id}"
+                )
+                nb_failed += 1
+            else:
+                nb_deleted += 1
+
+    logger.info(f"Done deleting zimcheck files from S3: {nb_deleted=}, {nb_failed=}")
```
benoit74: This is not zimcheck ^^

backend/src/cms_backend/shuttle/delete_uploaded_zims.py
```diff
@@ -0,0 +1,42 @@
...
+
+from cms_backend import logger
+from cms_backend.db.requested_task import get_requested_tasks
+from cms_backend.utils.s3 import get_kiwix_storage_client
+
+
+def delete_uploaded_zims(session: OrmSession):
+    """Delete uploaded zims for requested tasks"""
+
+    logger.info("Deleting zimcheck results from S3")
```
benoit74: This is not zimcheck ^^

backend/src/cms_backend/shuttle/delete_uploaded_zims.py
```diff
@@ -0,0 +1,42 @@
...
+    nb_deleted, nb_failed = 0, 0
+    omit_task_ids: list[UUID] = []
+
+    while True:
+        results = get_requested_tasks(
```
benoit74: I feel like this logic will forever try to delete S3 files, since there is nothing stored in DB to indicate that a given requested task has its file already deleted and not fetch it again on next invocation of this background task.

I also don't get why this is implemented in the shuttle, IMHO it would be better suited to run in the mill since it does not need to have access to storage server. It makes me realize we can probably move the logic to delete zimcheck results to the mill as well. The less we have on the shuttle, the better (storage server is already doing too many things, so the less we add to the shuttle which run on storage server, the better). Sorry for having overlooked this in the past. Or do I miss an important detail?

backend/src/cms_backend/api/routes/titles.py
```diff
@@ -345,6 +369,251 @@ def get_title_flavours(
     )
 
 
+class FileUploadRequest(BaseModel):
+    filename: NotEmptyString
+    filesize: int = Field(ge=0)
+    chunk_size: int = Field(ge=0)
```
benoit74: `part_size` would be more self-explicit IMHO

backend/src/cms_backend/api/routes/titles.py
```diff
@@ -345,6 +369,251 @@ def get_title_flavours(
...
+    upload_id: str | None = None
+
+
+@router.post(
+    "/{title_identifier}/upload/create",
```
benoit74: `create_or_resume` is more exact since this endpoint can be used to resume a previous upload if we have its upload ID, isn't it?

backend/src/cms_backend/api/routes/titles.py
```diff
@@ -345,6 +369,251 @@ def get_title_flavours(
...
+    """Generate presigned URLs for upload to S3 bucket"""
+    title = db_title.get_title(
+        session,
+        title_identifier=title_identifier,
+        accessible_collection_ids=accessible_collection_ids,
+    )
+    s3 = get_kiwix_storage_client(Context.zim_upload_s3_bucket_uri)
+    return generate_multipart_upload_presigned_urls(
+        s3,
+        key=f"uploads/titles/{title.name}/{uuid4()!s}.zim",
```
benoit74: Change convention to `cms_zim_uploads/{title.name}/{uuid4()!s}.zim` ; 1 level less in the hierarchy to simplify key and more explicit prefix for clarity.

backend/src/cms_backend/api/routes/titles.py
```diff
@@ -345,6 +369,251 @@ def get_title_flavours(
...
+    )
+
+
+class MultipartCompleteRequest(BaseModel):
+    upload_id: str
+    key: str
+    parts: list[PartEtag]
+
+
+class TaskCreateRequest(BaseModel):
```
benoit74: I don't get the distinction between `MultipartCompleteRequest` and `TaskCreateRequest`; I feel like `MultipartCompleteRequest` is a more explicit name but is unused. And `TitleUploadCompleteRequest` would more more accurate now.

dev/contrib/create_worker.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# Call it once to create a `test-worker`:
```
benoit74: Clarify this is for a **Zimfarm** worker

backend/src/cms_backend/api/routes/titles.py
elfkuzco: Initially, it had other fields nested in it like `collection_name` and `collection_path` but I decided to leave it nested in case we decided to add other fields in the future.

backend/src/cms_backend/api/routes/titles.py
elfkuzco: Yeah

backend/src/cms_backend/db/models.py
elfkuzco: We use it to later delete the zim after it's task has been deleted. This is the root of the bucket where the zims are uploaded. It is not the url to zimfarm which is a presigned url that expires after some days. I didn't want to fix it to an env variable because if env value changes, task to delete uploaded zims will fail  because they would assume those old requested zims were still uploaded to the same url. It is set from API context. Unless we are sure the root bucket will never change, I will proceed to delete it from DB

elfkuzco: Here's what it's set to in compose file:

```sh
      ZIM_UPLOAD_S3_BUCKET_URI: s3+https://s3.us-west-1.wasabisys.com/?keyId=xxxxxxxxxxxxxxxxxxxx&secretAccessKey=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx&bucketName=xxxxxxxxxxxxxxxxxx
```

backend/src/cms_backend/shuttle/delete_uploaded_zims.py
elfkuzco: moved shuttle logic to delete zimcheck files and this one to milll

elfkuzco:
applied the fixes and renamed the requested task to title upload across db, api routes and ui

backend/src/cms_backend/db/models.py
benoit74: It is safe to assume the bucket will never change, or we will easily handle the transition "by hand"

backend/src/cms_backend/api/routes/titles.py
benoit74: This looks like premature optimization. I agree we might at some point need it again ... or we might never need it again and constantly ask why we have two classes instead of one. Please keep only one.

backend/src/cms_backend/db/models.py
elfkuzco: applied

benoit74 (approved):
LGTM, thank you for this hard PR, a great step forward for the CMS! Glad it landed so nicely from a technical PoV.

Please rebase and let me know when it is ready to merge. I will merge, because I need to first setup S3 stuff and new env vars. And I will check everything works fine once deployed.

elfkuzco:
> Please rebase and let me know when it is ready to merge. I will merge, because I need to first setup S3 stuff and new env vars. And I will check everything works fine once deployed.

Rebased and converted to draft till you have set up env vars and S3-related stuff

